PR #1, opened 2019-05-31
Initial project scaffold

crdueck:
An initial project scaffold taking inspiration and READMEs from https://github.com/golang-standards/project-layout

Some relevant differences from the standard project layout:
* using go module vendoring
* no `pkg/` dir: we're not writing library code.
* no `cmd/` dir: we don't plan to generate multiple binaries


.travis.yml
```diff
@@ -0,0 +1 @@
+build/ci/travis.yml
```
crdueck: I blindly followed the recommendation of the standard project layout here (to symlink to `build/ci/*`), but is it worthwhile? We don't really need to support multiple CI configurations and it might surprise people.

PR #8, opened 2019-06-11
SQS Consumer for ship-it

rsgc-peterson-e:
VELO-1405

Uses function written in [PR #18](https://github.com/Wattpad/sqsconsumer/pull/18)


internal/sqs/sqs.go
```diff
@@ -0,0 +1,87 @@
...
+
+// Runs a preconfigured Consumer
+func RunConsumer(c *sqsconsumer.Consumer) {
+
+	numFetchers := 1
+
+	// set up a context which will gracefully cancel the worker on interrupt
+	fetchCtx, cancelFetch := context.WithCancel(context.Background())
+	term := make(chan os.Signal, 1)
+	signal.Notify(term, os.Interrupt, os.Kill)
```
crdueck: `signal.Notify(term, syscall.SIGINT, syscall.SIGTERM)`

rsgc-peterson-e: That worked. Thank you!

main.go
```diff
@@ -1,10 +1,77 @@
...
+	hist := dd.NewTiming("worker.time", 1.0).With("worker", "ship-it-worker", "queue", sqsConf.Name)
+
+	// AWS Setup
+	conf := &aws.Config{
+		Region: &envConf.AWSRegion,
+	}
+
+	s, err := session.NewSession(conf)
+	if err != nil {
+		fmt.Println(err)
```
crdueck: We should use the `go-kit` logger here. It has a `Fatalf` method that combines logging and exiting the process, since we can't continue if initializing these dependencies fails.

main.go
```diff
@@ -1,10 +1,72 @@
...
+	}
+
+	sqsConf, err := ecrconsumer.NewSQSConfig("kube-deploy-events.fifo", envConf)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// DataDog Setup
+	logger := kitlog.NewJSONLogger(kitlog.NewSyncWriter(os.Stdout))
+	dd := dogstatsd.New("wattpad.", logger)
```
crdueck: We should add our service name to this prefix, so we can distinguish the metrics from our service.

internal/ecrconsumer/sqs.go
```diff
@@ -0,0 +1,86 @@
...
+func NewSQSConfig(name string, env *Config) (*SQSConfig, error) {
+	return &SQSConfig{
+		Name:        name,
+		Region:      env.AWSRegion,
+		DatadogHost: env.DogstatsdHost,
+		DatadogPort: env.DogstatsdPort,
+	}, nil
+}
+
+func NewSQSConsumer(logger kitlog.Logger, dd *dogstatsd.Dogstatsd, hist metrics.Histogram, name string, svc sqsconsumer.SQSAPI) (*sqsconsumer.Consumer, error) {
```
crdueck: The `dd` client is unused and can be removed, the consumer only depends on the histogram.

internal/ecrconsumer/sqs.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// Metric Imports
+	kitlog "github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/metrics"
+	"github.com/go-kit/kit/metrics/dogstatsd"
+)
+
+// NOTE: Refactor to take Env struct
+
+type SQSConfig struct {
```
crdueck: This config type can be removed, the consumer only depends on the queue name which can be passed as a parameter.

main.go
```diff
@@ -28,7 +35,38 @@ func main() {
...
-	go http.ListenAndServe(":80", api.New())
+	// Load Environment Variables
+	envConf, err := FromEnv()
+	if err != nil {
+		log.Fatalf("Error getting environment variables %s", err)
+	}
+
+	dd := dogstatsd.New("wattpad.ship-it.", logger)
+	go dd.SendLoop(time.NewTicker(time.Second).C, "udp", envConf.DataDogAddress())
+	hist := dd.NewTiming("consumer.time", 1.0).With("consumer", "ship-it-consumer", "queue", envConf.QueueName)
```
crdueck: For consistency with other service's queue consumer metrics, this should be `worker.time`. Since this service will have multiple workers, we should use `worker:ecrconsumer` instead of `ship-it-consumer`.

main.go
```diff
@@ -28,7 +35,38 @@ func main() {
...
+	conf := &aws.Config{
+		Region: &envConf.AWSRegion,
+	}
+
+	s, err := session.NewSession(conf)
+	if err != nil {
+		log.Fatalf("Error opening AWS session %s", err)
+	}
+
+	svc := sqs.New(s)
```
crdueck: Let's inline the single use `svc` variable.

main.go
```diff
@@ -28,7 +35,38 @@ func main() {
 		cancel()
 	}()
 
-	go http.ListenAndServe(":80", api.New())
+	// Load Environment Variables
+	envConf, err := FromEnv()
+	if err != nil {
+		log.Fatalf("Error getting environment variables %s", err)
```
crdueck: These log messages should use the `go-kit` structured logger.

main.go
```diff
@@ -28,7 +35,38 @@ func main() {
...
+	}
+
+	dd := dogstatsd.New("wattpad.ship-it.", logger)
+	go dd.SendLoop(time.NewTicker(time.Second).C, "udp", envConf.DataDogAddress())
+	hist := dd.NewTiming("consumer.time", 1.0).With("consumer", "ship-it-consumer", "queue", envConf.QueueName)
+
+	// AWS Setup
+	conf := &aws.Config{
```
crdueck: Let's inline the single use `conf` variable.

main.go
```diff
@@ -28,7 +35,38 @@ func main() {
...
+	if err != nil {
+		log.Fatalf("Error opening AWS session %s", err)
+	}
+
+	svc := sqs.New(s)
+
+	// ECR SQS Consumer Setup
+	consumer, err := ecrconsumer.NewSQSConsumer(logger, hist, envConf.QueueName, svc)
+	if err != nil {
+		log.Println(err)
```
crdueck: This should use `log.Fatal`

main.go
```diff
@@ -2,18 +2,25 @@ package main
...
 	"ship-it/internal/api"
+	"ship-it/internal/ecrconsumer"
 	"syscall"
+	"time"
 
-	"github.com/go-kit/kit/log"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/sqs"
+	kitlog "github.com/go-kit/kit/log"
```
crdueck: We won't have to do this qualified import once we remove the usage of the stdlib `log` pkg.

internal/ecrconsumer/sqs.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"context"
+	"time"
+
+	"github.com/Wattpad/sqsconsumer"
+	"github.com/Wattpad/sqsconsumer/middleware"
+
+	// Metric Imports
+	kitlog "github.com/go-kit/kit/log"
```
crdueck: We can omit the qualified import, since the stdlib `log` package isn't in scope.

main.go
```diff
@@ -28,7 +35,38 @@ func main() {
...
 
-	go http.ListenAndServe(":80", api.New())
+	// Load Environment Variables
+	envConf, err := FromEnv()
+	if err != nil {
+		log.Fatalf("Error getting environment variables %s", err)
+	}
+
+	dd := dogstatsd.New("wattpad.ship-it.", logger)
+	go dd.SendLoop(time.NewTicker(time.Second).C, "udp", envConf.DataDogAddress())
```
crdueck: We can use `time.Tick(...)` here instead of `time.NewTicker(...).C`

main.go
```diff
@@ -28,7 +35,38 @@ func main() {
 		cancel()
 	}()
 
-	go http.ListenAndServe(":80", api.New())
+	// Load Environment Variables
```
crdueck: Generally, comments should add value by telling us why code does what it does, not what does the code do. We already know what the code does because we can read the code itself. There are exceptions if the code does something particularly tricky, but I don't think this is one of those cases. 

The func name `FromEnv` and the context around it, like the log message, already tells the reader that these lines of code load a config from env.

main.go
```diff
@@ -28,7 +35,38 @@ func main() {
...
+	envConf, err := FromEnv()
+	if err != nil {
+		log.Fatalf("Error getting environment variables %s", err)
+	}
+
+	dd := dogstatsd.New("wattpad.ship-it.", logger)
+	go dd.SendLoop(time.NewTicker(time.Second).C, "udp", envConf.DataDogAddress())
+	hist := dd.NewTiming("consumer.time", 1.0).With("consumer", "ship-it-consumer", "queue", envConf.QueueName)
+
+	// AWS Setup
```
crdueck: Does this comment add value or aid understanding of the code?

main.go
```diff
@@ -28,7 +35,38 @@ func main() {
...
+	}
+
+	s, err := session.NewSession(conf)
+	if err != nil {
+		log.Fatalf("Error opening AWS session %s", err)
+	}
+
+	svc := sqs.New(s)
+
+	// ECR SQS Consumer Setup
```
crdueck: Does this comment add value or aid understanding of the code?

internal/ecrconsumer/sqs.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Configure Middleware
+	track := dataDogTimeTracker(hist)
+	wrappedLogger := loggerMiddleware(logger)
+	handler := middleware.ApplyDecoratorsToHandler(processMessage, track, wrappedLogger)
+
+	// Create and return SQS consumer
```
crdueck: Does this comment add value or aid understanding of the code?

internal/ecrconsumer/sqs.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/Wattpad/sqsconsumer"
+	"github.com/Wattpad/sqsconsumer/middleware"
+
+	// Metric Imports
+	kitlog "github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/metrics"
+)
+
+func NewSQSConsumer(logger kitlog.Logger, hist metrics.Histogram, name string, svc sqsconsumer.SQSAPI) (*sqsconsumer.Consumer, error) {
+	// Create SQS service
```
crdueck: Does this comment add value or aid understanding of the code?

internal/ecrconsumer/sqs.go
```diff
@@ -0,0 +1,67 @@
+package ecrconsumer
+
+import (
+	"context"
+	"time"
+
+	"github.com/Wattpad/sqsconsumer"
+	"github.com/Wattpad/sqsconsumer/middleware"
+
+	// Metric Imports
```
crdueck: Does this comment add value or aid understanding of the code?

internal/ecrconsumer/sqs.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+func NewSQSConsumer(logger kitlog.Logger, hist metrics.Histogram, name string, svc sqsconsumer.SQSAPI) (*sqsconsumer.Consumer, error) {
+	// Create SQS service
+	service, err := sqsconsumer.NewSQSService(name, svc)
+	if err != nil {
+		return nil, err
+	}
+
+	// Configure Middleware
```
crdueck: Does this comment add value or aid understanding of the code?

config.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/kelseyhightower/envconfig"
+)
+
+type Config struct {
+	AWSRegion     string `envconfig:"AWS_REGION" required:"true"`
+	DogstatsdHost string `split_words:"true" required:"true"`
+	DogstatsdPort string `split_words:"true" default:"8125"`
+	QueueName     string `split_words:"true" required:"true"`
+	ServicePort   string `split_words:"true" required:"true"`
```
crdueck: Let's make this default to `"80"` rather than being required.

config.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+func (e *Config) DataDogAddress() string {
+	return net.JoinHostPort(e.DogstatsdHost, e.DogstatsdPort)
+}
+
+func FromEnv() (*Config, error) {
+	var env *Config
```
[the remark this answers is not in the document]
rsgc-peterson-e: I thought it needs to be a pointer because the function returns a pointer and in the case of an error I can return a `nil` pointer for the config. I did remove the `&` from the `envconfig.Process()` call while leaving the `*Config` type on `env`.

main.go
rsgc-peterson-e: `go-kit` does not seem to have an equivalent of `log.Fatal` so I used the normal `log.Log` function followed by an `os.Exit(1)` when the program fails to initialize its dependencies in `main()`.

config.go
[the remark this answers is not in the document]
crdueck: Either the `var env *Config` pointer type, or the `&env` indirection in the `Process` call was redundant. When we had both then `&env` is a `**Config`. Changing either works 👍

main.go
crdueck: Ah you're right, go-kit doesn't have that func. We should always be using the structured logger (see the logging section [here](https://github.com/Wattpad/highlander/blob/master/docs/microservice-checklist.md#expected-minimums)), so we can use `logger.Log` + `os.Exit(1)`.

main.go
```diff
@@ -29,6 +35,32 @@ func main() {
...
 
+	envConf, err := FromEnv()
+	if err != nil {
+		logger.Log("Error getting environment variables %s", err)
+		os.Exit(1)
+	}
+
+	dd := dogstatsd.New("wattpad.ship-it.", logger)
+	go dd.SendLoop(time.Tick(time.Second), "udp", envConf.DataDogAddress())
+	hist := dd.NewTiming("worker.time", 1.0).With("worker", "worker:ecrconsumer", "queue", envConf.QueueName)
```
crdueck: `With(key1, val1, ..., keyN, valN)` produces metrics tags like `key1:val1`, ..., `keyN:valN`, so this would create a metric tag `worker:worker:ecrconsumer`

crdueck:
I changed the base branch for this PR to `http-server-cancellation` yesterday so you could get a clean diff, but that branch has been merged into master now. You should edit this PR to be based off master before merging, it should rebase cleanly since nothing else has changed

PR #14, opened 2019-06-14
VELO-1422: stub /releases endpoint

crdueck:
The /releases endpoint will return a list of deployment details for
each service registered with ship-it.

---

To test locally:
```
$ make build
$ aws-vault exec wattpad -- make run
```

Then make a request to `$(docker-machine ip):8080/releases`. The business logic is stubbed out, so this returns `{"code":500,"message":"not implemented"}`

### DIFF
--- a/internal/models/deployment_detail.go
+++ b/internal/models/deployment_detail.go
@@ -1,0 +1,43 @@
+package models
+
+import (
+	"time"
+)
+
+type DeploymentDetail struct {
+	Name         string    `json:"name"`
+	Created      string    `json:"created"`
+	LastDeployed time.Time `json:"lastDeployed"`
+	Owner        struct {
+		Team  string `json:"team"`
+		Slack string `json:"slack"`
+	} `json:"owner"`
+	AutoDeploy bool `json:"autoDeploy"`
+	Code       struct {
+		Github string `json:"github"`
+		Ref    string `json:"ref"`
+	} `json:"code"`
+	Build struct {
+		Travis string `json:"travis"`
+	} `json:"build"`
+	Monitoring struct {
+		Datadog struct {
+			Dashboard string `json:"dashboard"`
+			Monitors  string `json:"monitors"`
+		} `json:"datadog"`
+		Sumologic string `json:"sumologic"`
+	} `json:"monitoring"`
+	Artifacts struct {
+		Docker struct {
+			Image string `json:"image"`
+			Tag   string `json:"tag"`
+		} `json:"docker"`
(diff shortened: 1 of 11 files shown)


internal/models/deployment_detail.go
```diff
@@ -0,0 +1,43 @@
+package models
+
+import (
+	"time"
+)
+
+type DeploymentDetail struct {
```
crdueck: Open to suggestions on this name. I didn't want to call it `Deployment`, since it has a `Deployment` field.

internal/models/deployment_detail.go
```diff
@@ -0,0 +1,43 @@
...
+		Docker struct {
+			Image string `json:"image"`
+			Tag   string `json:"tag"`
+		} `json:"docker"`
+		Chart struct {
+			Path    string `json:"path"`
+			Version string `json:"version"`
+		} `json:"chart"`
+	} `json:"artifacts"`
+	Deployment struct {
```
cfmtam: What do you think about renaming this to `Release`?

crdueck: I think that'd make sense. We were talking about listing the k8s resources in this section, and those resources belong to the helm release rather than a k8s deployment. 

The terminology is starting to get confusing though, we would have a `DeploymentDetail` that contains details about a helm release, which owns a k8s `Deployment` resource. Would we want to rename this struct to `ReleaseDetail`, and change the endpoint to `/releases`?

cfmtam: I like changing the endpoint to `/releases` and renaming this struct to `Release` or `ReleaseDetail`.

For the `Deployment` struct, is `status` the only thing in there (since resources will be a separate endpoint)? Is this status from ship-it, helm, or k8s? I would kind of argue for simplifying this struct into just a `status` string.

crdueck: True, I think we can have just the `Status` field in this response, and move the release resources to another response type. I'm imagining a future API like

```
GET /releases
GET /releases/{name}
GET /releases/{name}/resources 
```

We avoid expensively listing all the resources for every release, but if someone clicks into a detailed view of a release in the UI, we can return the resources for that particular release.

crdueck:
Renamed some things and updated the schema doc, another R?

jrhouston:
Can we set up golint? Lot of exported things here without any comments. I just have it lint in my editor so it pesters me to add the comments.

crdueck:
I'm going to fixup merge conflict and address linter complaints, then re-assign for review.

PR #21, opened 2019-06-17
Add Github integration to find TravisCI URL from Checks API

jrhouston:
Nothing to plumb this through to yet, but adding in the module + config


internal/integrations/github/github_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/h2non/gock.v1"
+)
+
+func TestGetTravisCIBuildURLForRef_Success(t *testing.T) {
+	defer gock.Off()
+
+	mockResponse, _ := ioutil.ReadFile("testdata/check_runs_response_success.json")
+	gock.New("https://api.github.com").
```
jrhouston: https://github.com/google/go-github didn't seem to have a way of stubbing a response at a higher level so this just intercepts the HTTP call to the GitHub Checks API and injects a response that I `curl`'d myself from the API.

internal/integrations/github/github.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+// GitHub handles integrations with the GitHub API
+type GitHub struct {
+	Org    string
+	Client *github.Client
+}
+
+// New creates a new GitHub integration
+func New(org string, accessToken string) (*GitHub, error) {
```
crdueck: This func cannot fail, so there's no need to return an error.

internal/integrations/github/github.go
```diff
@@ -0,0 +1,47 @@
...
+
+	return &GitHub{
+		Client: client,
+		Org:    org,
+	}, nil
+}
+
+// GetTravisCIBuildURLForRef uses the Checks API to find the URL to the Travis build for the specified ref
+func (github *GitHub) GetTravisCIBuildURLForRef(repo string, ref string) (string, error) {
+	ctx := context.Background()
```
crdueck: This context should be provided as a parameter. Once we have the HTTP controller code in place, the request's context will get plumbed through to this method.

internal/integrations/github/github.go
```diff
@@ -0,0 +1,47 @@
...
+
+	checks, _, err := github.Client.Checks.ListCheckRunsForRef(ctx, github.Org, repo, ref, nil)
+
+	if err != nil {
+		return "", err
+	}
+
+	for _, checkRun := range checks.CheckRuns {
+		if *checkRun.App.Name == "Travis CI" {
```
crdueck: We should use the `GetX` convenience methods here (ie.`checkRun.GetApp().GetName()`). These methods handle the potential nil pointers by returning the zero value instead of panic-ing

internal/integrations/github/github.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"context"
+	"errors"
+
+	"github.com/google/go-github/v26/github"
+	"golang.org/x/oauth2"
+)
+
+// GitHub handles integrations with the GitHub API
+type GitHub struct {
```
crdueck: bike-shedding: I've seen "github the corporation" stylized as GitHub, but "github the website/service" is typically stylized as Github. I'd prefer `Github` but don't feel too strongly either way, as long as it's consistent. Specifically, the config fields added in this PR are stylized as `Github*` which doesn't match this struct.

internal/integrations/github/github.go
```diff
@@ -0,0 +1,47 @@
...
+	client := github.NewClient(tc)
+
+	return &GitHub{
+		Client: client,
+		Org:    org,
+	}, nil
+}
+
+// GetTravisCIBuildURLForRef uses the Checks API to find the URL to the Travis build for the specified ref
+func (github *GitHub) GetTravisCIBuildURLForRef(repo string, ref string) (string, error) {
```
crdueck: The `github` method receiver shadows the `go-github/github` package symbol. Usually method receiver vars are abbreviated because their scope is limited and the context they're used in is well understood. Should we use an abbreviated method receiver here?

internal/integrations/github/github_test.go
crdueck: Another option would be to use an interface to decouple our `Github` struct type from the `go-github/github.Client`, and then mock out that dependency in these test cases.

internal/integrations/github/github_test.go
```diff
@@ -0,0 +1,52 @@
...
+	mockResponse, _ := ioutil.ReadFile("testdata/check_runs_response_empty.json")
+	gock.New("https://api.github.com").
+		Get("/repos/Wattpad/highlander/commits/2c76895cdb9f3ff5100ecf93a7a6c6747aaeda8c/check-runs").
+		Reply(200).
+		JSON(mockResponse)
+
+	github, _ := New("Wattpad", "fake-access-token")
+	_, err := github.GetTravisCIBuildURLForRef("highlander", "2c76895cdb9f3ff5100ecf93a7a6c6747aaeda8c")
+
+	assert.Equal(t, "could not find a TravisCI build for specified ref", err.Error())
```
crdueck: Using [`assert.EqualError`](https://godoc.org/github.com/stretchr/testify/assert#EqualError) would be slightly cleaner, but I don't like that this test will fail if the message of the error returned by `GetTravisCIBuildURLForRef` ever changes. 

Tests should verify a behaviour of the implementation (return an error when the response contains no check-runs), not the implementation details of that behaviour (the specific string in the error message)

My suggestion is to define a `var errTravisCIBuildNotFound`, then use that var in both the test cases and the code being tested.

internal/integrations/github/github_test.go
```diff
@@ -0,0 +1,52 @@
...
+	defer gock.Off()
+
+	gock.New("https://api.github.com").
+		Get("/repos/Wattpad/highlander/commits/2c76895cdb9f3ff5100ecf93a7a6c6747aaeda8c/check-runs").
+		Reply(500)
+
+	github, _ := New("Wattpad", "fake-access-token")
+	_, err := github.GetTravisCIBuildURLForRef("highlander", "2c76895cdb9f3ff5100ecf93a7a6c6747aaeda8c")
+
+	assert.NotEmpty(t, err)
```
crdueck: https://godoc.org/github.com/stretchr/testify/assert#Error seems more appropriate

internal/integrations/github/github_test.go
```diff
@@ -0,0 +1,80 @@
+package github
```
jrhouston: Feel like the test suite got a lot more confusing.

internal/integrations/github/github.go
```diff
@@ -0,0 +1,53 @@
...
+
+// ChecksService wraps github.ChecksService
+type ChecksService interface {
+	ListCheckRunsForRef(context.Context, string, string, string, *github.ListCheckRunsOptions) (*github.ListCheckRunsResults, *github.Response, error)
+}
+
+// Github handles integrations with the Github API
+type Github struct {
+	Org    string
+	Client *github.Client
```
crdueck: The `Client` field is unused, should it be removed?

internal/integrations/github/github_test.go
```diff
@@ -0,0 +1,78 @@
...
+	g := New(ctx, "Wattpad", "fake-access-token")
+
+	fakeError := errors.New("fake")
+	m := new(MockChecksService)
+	m.On("ListCheckRunsForRef").Return(nil, nil, fakeError)
+	g.Checks = m
+
+	_, err := g.GetTravisCIBuildURLForRef(ctx, "highlander", "master")
+
+	assert.EqualError(t, fakeError.Error(), err)
```
crdueck: It looks like the behaviour we want to test here is that when the `ChecksService` returns an error, then `GetTravisCIBuildURLForRef` should return an error. However, this test case would break if `GetTravisCIBuildURLForRef` added context to the checks error with `errors.Wrap`, even though that's a valid way to implement the desired behaviour.

Since the error value is arbitrary (it's defined by the test case), I would prefer just `assert.Error(t, err)`. If it's particularly important that the `GetTravisCIBuildURLForRef` error is related to the original checks error, then we could do `assert.Equal(t, fakeError, errors.Cause(err))`.

internal/integrations/github/github_test.go
crdueck: I don't think it's too confusing. Defining the `github.ListCheckRunResults` values by hand is a bit verbose, but we only do that in one place. A `newCheckRun(detailsURL, app string) *github.CheckRun` test helper could help there, but for now I don't think it's necessary.

internal/integrations/github/github_test.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func TestGetTravisCIBuildURLForRef_Success(t *testing.T) {
+	ctx := context.Background()
+	g := New(ctx, "Wattpad", "fake-access-token")
+
+	m := new(MockChecksService)
+	m.On("ListCheckRunsForRef").Return(&github.ListCheckRunsResults{
+		CheckRuns: []*github.CheckRun{
+			&github.CheckRun{
```
crdueck: Doesn't make much difference, but this `&github.CheckRun` syntax can be omitted when defining non-composite-literal types (slices and maps). Since the type of an element of a `[]*github.CheckRun` is implied, I like removing the extraneous syntax.

internal/integrations/github/github.go
```diff
@@ -0,0 +1,53 @@
...
+	Checks ChecksService
+}
+
+// New creates a new GitHub integration
+func New(ctx context.Context, org string, accessToken string) *Github {
```
crdueck: It would make testing slightly more straightforward if this constructor func depended directly on the `ChecksService` interface, ie. `New(checks ChecksService, org string) *Github`. 

The downside is that the caller now needs to know about the oauth and github pkgs so it can provide a client. I'm not sure whether slightly cleaner tests are worth losing that encapsulation. It might become easier to decide as more things are added to this github pkg, so I'd be okay punting on it.

jrhouston: Yeah I considered doing that but I want to keep those details hidden for now.

internal/integrations/github/github_test.go
```diff
@@ -2,55 +2,79 @@ package github
...
-	"gopkg.in/h2non/gock.v1"
+	"github.com/stretchr/testify/mock"
 )
 
+type MockChecksService struct {
+	mock.Mock
+}
+
+func (m *MockChecksService) ListCheckRunsForRef(context context.Context, owner string, repo string, ref string, opt *github.ListCheckRunsOptions) (*github.ListCheckRunsResults, *github.Response, error) {
+	args := m.Called()
```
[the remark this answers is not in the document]
jrhouston: I'm not so interested in having tests that make assertions about what internal calls are made by a function.

internal/integrations/github/github.go
crdueck: `Client` field init was removed from `New`, but the field is still declared here.

PR #22, opened 2019-06-17
Consume the API for querying registered services on the UI

rsgc-peterson-e:
VELO-1428

A change that must be made is to point the dashboard to the correct web address for the API as it currently points to a local test server on my machine.


web/src/components/SelectionDialog.js
```diff
@@ -21,7 +22,19 @@ class SelectionDialog extends React.Component {
     this.setState({ open: true })
   }
 
+  getMirandaURL() {
+    var url = urljoin('https://github.com/Wattpad/highlander/tree/master/k8s/charts/services', this.props.serviceName)
```
crdueck: We should be able to configure the base urls for our "source code repo" and "chart repo", without having them specifically reference highlander and miranda.

web/src/components/SelectionDialog.js
```diff
@@ -21,7 +22,19 @@ class SelectionDialog extends React.Component {
...
+    return url
+  }
+
+  getHighlanderURL() {
+    var url = urljoin('https://github.com/Wattpad/highlander/tree/', this.props.gitref, this.props.highlanderPath)
```
crdueck: I would avoid naming this single use var and just `return urljoin(...)` directly

rsgc-peterson-e: My bad, it was initially configured that way and then I moved it into its own var when debugging an error as it looked like `urljoin()` was not returning a string.

web/src/components/GridCell.js
```diff
@@ -0,0 +1,114 @@
...
+import SelectionDialog from './SelectionDialog'
+
+import TimePassed from '../assets/time_passed.png'
+import SlackIcon from '../assets/slack_icon.png'
+import DockerIcon from '../assets/docker_icon.png'
+import StatusChip from './StatusChip';
+import SlackInfoDialog from './SlackInfoDialog';
+import ImageInfoDialog from './ImageInfoDialog';
+
+const imgAlt = "not found"
```
jrhouston: Not sure what this is for, we can just omit the alt tag from images because we're not trying to be accessible.

web/src/components/GridCell.js
```diff
@@ -0,0 +1,114 @@
...
+                }
+                <IconButton onClick={this.dockerClicked}>
+                  <img src={DockerIcon} width="32" height="32" alt={imgAlt} />
+                </IconButton>
+                {
+                  this.state.imageInfo ? <ImageInfoDialog open={this.state.imageInfo} handleClose={this.dockerClosed} docker={this.props.SingleGridCellData.artifacts.docker} /> : null
+                }
+              </div>
+              <div>
+                <IconButton onClick={this.cellClick.bind(this)}>
```
jrhouston: Why did we have to bind here and not use the `cellClick = () =>{` syntax in the method?

web/src/components/ImageInfoDialog.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
+import Dialog from '@material-ui/core/Dialog'
+import { DialogContent, DialogActions, Button, DialogTitle, Typography } from '@material-ui/core';
+
+class ImageInfoDialog extends React.Component {
+
```
jrhouston: Superfluous newline

jrhouston:
Mostly nitpicking :trollface:

web/src/components/ExpandedDetail.js
```diff
@@ -53,16 +53,17 @@ class ExpandedDetail extends React.Component {
   constructor(props) {
     super(props)
     this.state = {
-      podList: false,
-      resourceList: false,
+      podsVisible: false,
+      resourceVisible: false,
```
crdueck: This should be pluralized as well, matching `podsVisible`

web/src/components/ExpandedDetail.js
```diff
@@ -26,100 +30,187 @@ const theme = createMuiTheme({
...
-            <div className="right-padded">
-              <img src={HelmIcon} width="32" height="32" alt={imgAlt} />
-            </div>
-            <div>
-              <Typography variant="h5">ChartVersion</Typography>
-            </div>
-          </div>
+  podClick = (event) => {
+    this.setState({podList: !this.state.podList})
+    var api = urljoin('https://' + window.location.host, '/releases/', this.props.data.name, 'resources') // point to local IP for testing
```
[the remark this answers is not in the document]
crdueck: After learning the difference between `let` vs. `var`, the scoping behaviour of `let` is definitely less surprising and makes sense as the preferred default.

web/src/components/SelectionDialog.js
```diff
@@ -22,19 +22,18 @@ class SelectionDialog extends React.Component {
...
-    return url
+  getSourceURL(url) {
+    return urljoin(url, this.props.gitref, this.props.highlanderPath)
   }
 
   render() {
-    var highlanderURL = this.getHighlanderURL()
-    var mirandaURL = this.getMirandaURL()
+    var highlanderURL = this.getSourceURL('https://github.com/Wattpad/highlander/tree/')
+    var mirandaURL = this.getChartURL('https://github.com/Wattpad/highlander/tree/master/k8s/charts/services')
```
crdueck: As we discussed offline, these URLs should come directly from the `/releases` response body

web/src/App.js
```diff
@@ -2,62 +2,41 @@ import React from 'react'
...
-      chart: {
-        path: "https://charts.wattpadhq.com/homeslice",
-        version: "1.2.3"
-      }
-    },
-    deployment: {
-      status: "deploying", // deploying, deployed, rolled_back, failed
-      // NOTE resources separate endpoint
-    }
+const urljoin = require('url-join')
```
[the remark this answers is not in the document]
rsgc-peterson-e: It was a standard node package so I was unsure if the ES6 import would be supported. Fixed.

web/src/components/GridCell.js
rsgc-peterson-e: Agreed. The tag is only present so the linter passes and so there is no compiler warnings.

web/src/components/GridCell.js
rsgc-peterson-e: It was not for any reason, just some legacy code from the component I originally forked. Fixed.

web/src/components/ExpandedDetail.js
```diff
@@ -65,15 +65,8 @@ class ExpandedDetail extends React.Component {
...
-      var pods = []
-      var resources = []
-      for (var i = 0; i < response.data.length; i++) {
-        if (response.data[i].kind === "Pod") {
-          pods.push(response.data[i])
-        } else {
-          resources.push(response.data[i])
-        }
-      }
+      let pods = response.data.filter((d) => d.kind === 'Pod')
```
jrhouston: We could actually just put these lines inside the setState call below.

rsgc-peterson-e: Fixed.

web/src/App.js
```diff
@@ -2,62 +2,39 @@ import React from 'react'
...
-      chart: {
-        path: "https://charts.wattpadhq.com/homeslice",
-        version: "1.2.3"
-      }
-    },
-    deployment: {
-      status: "deploying", // deploying, deployed, rolled_back, failed
-      // NOTE resources separate endpoint
-    }
+const API_ADDRESS = urljoin('https://' + window.location.hostname, 'releases') // Point to local IP for testing
```
crdueck: After the recent changes to the API routing, this would be `/api/releases`

web/src/components/ExpandedDetail.js
```diff
@@ -26,100 +31,178 @@ const theme = createMuiTheme({
...
-            <div className="right-padded">
-              <img src={HelmIcon} width="32" height="32" alt={imgAlt} />
-            </div>
-            <div>
-              <Typography variant="h5">ChartVersion</Typography>
-            </div>
-          </div>
+  podClick = () => {
+    this.setState({podsVisible: !this.state.podsVisible})
+    let API_ADDRESS = urljoin('https://' + window.location.host, '/releases/', this.props.data.name, 'resources') // point to local IP for testing
```
crdueck: It's smelly that we build the API address in 2 places. It'd be cleaner if the root `https://${window.location.host}/api` was only defined once, and then the `ExpandedDetail` & `App` components could append their specific routes on top of it.

Is the `API_ADDRESS` const from `App.js` in scope here? If so, we could have that const be only the root, instead of the root + `/releases` endpoint, and share it between the components.

web/src/components/TileGrid.js
```diff
@@ -3,131 +3,25 @@
...
-      gridData: JSON.parse(this.props.gridData)
+      gridData: this.props.gridData
     }
   }
 
   findIdNode(target) { // takes event target and returns the id (written to make sure the expanded detail renders from the correct reference point as in the original component each tile on the grid was just an img tag)
-    var t = target
-    var found = false
+    let t = target
+    let found = false
```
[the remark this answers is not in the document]
rsgc-peterson-e: Fixed.

web/src/components/ExpandedDetail.js
rsgc-peterson-e: The API address is already fed into the Grid component as a prop. I will make use of this and feed it from the grid to the detail when it is rendered.

PR #23, opened 2019-06-17
Add makefile target to autogenerate jsonschema docs

crdueck:
The docs are generated directly from a golang struct, which means
developers will have to remember to add their struct to
`jsonschema/main.go` every time. We could consider adding a check that
ensures each model in `internal/models/XYZ.go` has a corresponding
`api/XYZ.json`.

---

This still needs a bit of polish (also won't pass CI until I rebase in the `Release` struct from https://github.com/Wattpad/ship-it/pull/14), but I had some questions I wanted to get opinions on:

* I like autogenerating the schemas directly from the Go code. The alternative is generating them from example JSON data, but that's one more thing to maintain in addition to the structs. Any thoughts on this?
* Not sure whether to add the jsonschema pkg to our vendored deps (adding to CI time) and use plain `go run` in the Makefile vs. build the docs in a container and keep our vendored deps to a minimum.
* felt weird to add this to the project's `cmd` dir, is there a better way to manage it? Would we want to separate it as a generally useful tool?


jrhouston:
Yeah I wouldn't put this in the cmd dir, that would be where code for the CLI would live.

jrhouston:
The [standard project layout](https://github.com/golang-standards/project-layout/tree/master/tools) proposes a `tools` directory at the top level of the repo - I'd go with that.

jrhouston:
I think having this in our vendored deps is fine for now. We can optimize later.

jrhouston:
I kind of want the CI to block if there was a diff between the doc and what's being checked in. Maybe later autogenerating the schema can be a consequence of merging to master?

crdueck:
We could create a pre-push git hook that runs the autogen script if files in `internal/models` changes.

tools/jsonschema/main.go
```diff
@@ -0,0 +1,39 @@
...
+	encoder.SetIndent("", "  ")
+
+	return encoder.Encode(jsonschema.Reflect(v))
+}
+
+func main() {
+	// NOTE: After creating a new JSON model type, it must be added to this
+	// map in order to auto-generate JSON schema documents.
+	types := map[string]interface{}{
+		"release": models.Release{},
```
jrhouston: Seems like it might be possible to automate this using the `go/importer` package to discover the models that are defined.

jrhouston: I started looking into it and went down a rabbit hole, something for another day.

crdueck: Took a quick look, that `importer` pkg seems pretty close. If we could enumerate the exported symbols of a package and filter by struct types only, I think that'd get us there.

internal/models/release.go
```diff
@@ -5,37 +5,53 @@ import (
...
+	Owner        Owner      `json:"owner" jsonschema:"description=Ownership and contact information"`
+	AutoDeploy   bool       `json:"autoDeploy" jsonschema:"description=The state of the release's auto-deployment option"`
+	Code         SourceCode `json:"code" jsonschema:"description=The repository and branch ref of the release's source code"`
+	Build        Build      `json:"build" jsonschema:"description=The CI build page of current release,required=true"`
+	Monitoring   Monitoring `json:"monitoring" jsonschema:"description=The monitoring resources for the release"`
+	Artifacts    Artifacts  `json:"artifacts" jsonschema:"description=The build artifacts of the release"`
+	Status       string     `json:"status" jsonschema:"description=The status of the release,example=deployed,example=failed,example=pending_rollback,example=pending_install,example=pending_upgrade"`
+}
+
+type Owner struct {
```
crdueck: These extra structs add a lot of bloat to the `models` pkg unfortunately. They either have to be public or we need to write a bunch of constructors/setters to allow other packages to construct `Release` values without having access to the internal types.

crdueck: I guess it would be really awkward to construct a `Release` with the anonymous structs as well, we just hadn't gotten that far yet.

Since the `Release` depends on some coming from k8s, github, etc. It might be easier to populate it incrementally instead of all at once (although it's cleaner to do it all at once, just more awkward), so we might want to have a fluent API that allows it to be constructed incrementally, like

```
var rls Release
rls.WithName("name").WithCreated(time.Now()).WithStatus("deployed") // etc
```

crdueck: Thinking about it some more, we can just have public fields with private types, and set fields directly without any fluent/builder API.

jrhouston:
Need to fix conflicts

PR #24, opened 2019-06-18
Implement business logic to query k8s api for /releases

cfmtam:
Implement business logic to query k8s api for registered services and results in the `/releases` endpoint. 

For now, we're just using ConfigMaps and just populating the `name` and `created` fields.

- Add k8s.io/api, client-go, and api-machinery packages
    
VELO-1437


internal/integrations/k8s/k8s.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (k *K8sClient) ListAll() (*v1.ConfigMapList, error) {
+	return k.ConfigMaps(v1.NamespaceAll).List(metav1.ListOptions{})
+}
+
+type K8sClient struct {
+	kv1.CoreV1Interface
+}
+
+func NewK8sClient() (K8sInterface, error) {
```
crdueck: We should avoid stuttering (`k8s.NewK8sClient`) and name the constructor `New` instead.

cfmtam:
@crdueck sorry, I rebase/squashed it and it got rid of the PR approval

PR #25, opened 2019-06-19
Make stable identifiers for API Data Dog metrics

rsgc-peterson-e:
VELO-1442


internal/api/middleware/timer.go
```diff
@@ -0,0 +1,46 @@
...
+	"strings"
+
+	"github.com/go-chi/chi"
+	"github.com/go-kit/kit/metrics"
+	"github.com/go-kit/kit/metrics/dogstatsd"
+)
+
+func getHist(r *http.Request, t *dogstatsd.Timing) metrics.Histogram {
+	str := strings.ReplaceAll(chi.RouteContext(r.Context()).RoutePattern(), "/", ".")
+	fmt.Println(str)
```
crdueck: we can remove the debug println

internal/api/router.go
```diff
@@ -18,10 +20,11 @@ type Service interface {
 }
 
 // New returns an 'http.Handler' that serves the ship-it API.
-func New(s Service) http.Handler {
+func New(s *service.Service) http.Handler {
```
crdueck: The histogram should be provided as a dependency to the API handler, since that's where it's used, rather than the service.

internal/api/middleware/timer.go
```diff
@@ -0,0 +1,46 @@
...
+
+func Timer(t *dogstatsd.Timing) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(
+			func(w http.ResponseWriter, r *http.Request) {
+				next.ServeHTTP(w, r)
+				h := getHist(r, t)
+				defer func(t0 time.Time) {
+					h.Observe(millisecondsSince(t0))
+				}(time.Now())
```
crdueck: We need to get the initial time value _before_ calling the next handler.

internal/api/middleware/timer.go
```diff
@@ -0,0 +1,46 @@
...
+	fmt.Println(str)
+	regx := regexp.MustCompile(`[{}]`)
+	if regx.MatchString(str) {
+		str = regx.ReplaceAllString(str, "")
+	}
+
+	str = strings.Replace(str, ".", "", 1)
+
+	fmt.Println(str)
+	return t.With(str, "method", r.Method)
```
crdueck: The metric label tags must always be provided in pairs, but only 3 are provided here. This is should look like `t.With("endpoint", str, "method", r.Method)`

internal/api/middleware/timer.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/go-chi/chi"
+	"github.com/go-kit/kit/metrics"
+	"github.com/go-kit/kit/metrics/dogstatsd"
+)
+
+func getHist(r *http.Request, t *dogstatsd.Timing) metrics.Histogram {
+	str := strings.ReplaceAll(chi.RouteContext(r.Context()).RoutePattern(), "/", ".")
+	regx := regexp.MustCompile(`[{}]`)
+	if regx.MatchString(str) {
```
crdueck: I don't think we need to check for matches first. If there are no matches, then `ReplaceAllString` does nothing anyways.

internal/api/middleware/timer.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"regexp"
+	"strings"
+
+	"github.com/go-chi/chi"
+	"github.com/go-kit/kit/metrics"
+	"github.com/go-kit/kit/metrics/dogstatsd"
+)
+
+func getHist(r *http.Request, t *dogstatsd.Timing) metrics.Histogram {
```
crdueck: As discussed offline, this function does too much. I'd suggest writing a single responsibility function that gets and formats the stable endpoint id, and move the rest of the responsibilities (getting the request method, adding the metric labels) to the middleware func.

internal/api/middleware/timer.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/go-chi/chi"
+	"github.com/go-kit/kit/metrics"
+)
+
+func getIdentifier(ctx context.Context) string {
+	str := strings.ReplaceAll(chi.RouteContext(ctx).RoutePattern(), "/", ".")
+	regx := regexp.MustCompile(`[{}]`)
+
+	str = regx.ReplaceAllString(str, "")
```
crdueck: We can use a [`strings.Replacer`](https://golang.org/pkg/strings/#Replacer) to handle all of the string substitutions in a single pass, and it's more efficient than regex: https://play.golang.org/p/bZ1J__BLLQR

internal/api/middleware/timer.go
```diff
@@ -0,0 +1,39 @@
...
+	str = strings.Replace(str, ".", "", 1)
```
crdueck: These last two lines can be written `return str[1:]`, taking all but the first (index 0) element of the slice.

internal/api/middleware/timer.go
```diff
@@ -0,0 +1,39 @@
...
+	"net/http"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/go-chi/chi"
+	"github.com/go-kit/kit/metrics"
+)
+
+func getIdentifier(ctx context.Context) string {
```
crdueck: This is now a nicely encapsulated bit of logic that should be tested. It might be difficult to create a `chi.Context` test value though, I don't see a way for us to put route patterns into it. 

So, I would see if you can find a clean way to create a test context, but if not, then we probably change this func to `getIdentifier(string) string` and have the `Timer` middleware call `RoutePattern()` instead.

internal/api/middleware/timer.go
crdueck: This one is still a thing. The cute `defer` trick likely won't work anymore, since we need a value created before calling the next handler (the initial time) and a value created after calling the next handler (the request context).

rsgc-peterson-e: If I get an initial time before calling the next handler does making use of `defer` after the handler call still work?

```
t0 := time.Now()
next.ServeHTTP(w, r)
id := getIdentifier(r.Context())
defer func(t0 time.Time) {
	h.With("endpoint", id, "method", r.Method).Observe(millisecondsSince(t0))
}(t0)
```

internal/api/middleware/timer.go
rsgc-peterson-e: The timer middleware now calls `RoutePattern` so `getIdentifier` has the suggested signature.

PR #27, opened 2019-06-20
Put go linting and testing into one job

rsgc-peterson-e:
VELO-1445

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,8 @@ jobs:
         - go test ./...
     
     - stage: Test
-      name: JSLint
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -24,17 +24,15 @@ jobs:
       script: make build
 
     - stage: Test
-      name: GoLint
+      name: GoLint & Test
```
crdueck: I'd just call this job `Go`, and the js one `Javascript`. Since they're already part of the `Test` stage, it looks a bit redundant:

Screen Shot 2019-06-20 at 12 37 05 PM

PR #28, opened 2019-06-20
Implement Custom Resource Editing VELO-1453

rsgc-peterson-e:
VELO-1453


rsgc-peterson-e:
This PR adds functionality to update images in custom resources given that the YAML is in the following format:

```
apiVersion: helmreleases.k8s.wattpad.com/v1alpha1
kind: HelmRelease
metadata:
  name: example-microservice
spec:
  releaseName: example-release

  chart:
    repository: wattpad.s3.amazonaws.com/helm-charts
    path: microservice
    revision: HEAD

  values:
    image:
      repository: dkr.ecr.amazonaws.com/foo
      tag: foo

    iamRoleName: foo
    serviceAccountName: foo

    autoscaler:
      minPods: 1
      maxPods: 2
      targetCPUUtilizationPercent: 80
```

An example of an image update is given below:

```
client := ecrconsumer.NewGitHub(token, context.Background(), "rsgc-peterson-e", "go-github-test")

img, err := ecrconsumer.LoadImage("loki", client)
if err != nil {
	fmt.Println(err)
	return
}
fmt.Println(img)
img.Tag = "foo"
err = img.Update(client)
if err != nil {
	fmt.Println(err)
	return
}
```

internal/ecrconsumer/image_values.go
```diff
@@ -0,0 +1,84 @@
...
+
+	return &Image{
+		service:    serviceName,
+		Registry:   registry,
+		Repository: repository,
+		Tag:        customResource.Spec.Values.Image.Tag,
+	}, nil
+}
+
+func (i *Image) Update(client GitCommands) error {
```
crdueck: I recommend to avoid smart models that implement business logic like this `Update` method. The `Image` model should be a dumb & simple data type and higher-level behaviours should be implemented elsewhere, like the ecrconsumer's message handler.

If we decompose this function into the pure/impure parts, then we could imagine writing pseudo-code like this in the ecrconsumer:

```
img := getDesiredImageFromECREvent(event)
rls := getHelmReleaseCRFile(releaseName)
putHelmReleaseCRFile(rls.WithImage(img))
```

internal/ecrconsumer/git.go
```diff
@@ -0,0 +1,122 @@
...
+	"golang.org/x/oauth2"
+)
+
+type GitCommands interface {
+	UpdateFile(msg string, branch string, path string, fileContent []byte) (*github.RepositoryContentResponse, error)
+	DownloadFile(branch string, path string) ([]byte, error)
+}
+
+type GitHub struct {
+	Client       *github.Client
```
crdueck: This field should be private, since we want callers to use the methods on `Github` instead of accessing the `*github.Client` directly.

internal/ecrconsumer/git.go
```diff
@@ -0,0 +1,122 @@
...
+	Respository  string
+}
+
+func NewGitHub(token string, ctx context.Context, org string, repo string) GitHub {
+	tokenSource := oauth2.StaticTokenSource(
+		&oauth2.Token{AccessToken: token},
+	)
+	tokenClient := oauth2.NewClient(ctx, tokenSource)
+
+	client := GitHub{
```
crdueck: let's inline this single use variable.

internal/ecrconsumer/git.go
```diff
@@ -0,0 +1,122 @@
...
+		if err := os.MkdirAll(filepath.Dir(path), os.ModePerm); err != nil { // create all folders in the relative path
+			return err
+		}
+
+		fileContent, err := file.GetContent()
+		if err != nil {
+			return err
+		}
+
+		if err := ioutil.WriteFile(path, []byte(fileContent), 0700); err != nil { // write the files in the corrects spots in the directory tree
```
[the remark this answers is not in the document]
rsgc-peterson-e: Fixed.

internal/ecrconsumer/cr.go
```diff
@@ -0,0 +1,31 @@
+package ecrconsumer
+
+// Definition of Custom Resource YAML
+type CRYaml struct {
```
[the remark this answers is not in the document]
rsgc-peterson-e: I have renamed the struct to `HelmRelease` but left it in the current package until we have greater consensus on where to put it.

internal/ecrconsumer/cr.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+	Spec struct {
+		ReleaseName string `yaml:"releaseName"`
+		Chart       struct {
+			Repository string `yaml:"repository"`
+			Path       string `yaml:"path"`
+			Revision   string `yaml:"revision"`
+		}
+		Values struct {
+			Image struct {
```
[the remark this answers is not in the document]
rsgc-peterson-e: I added a function to parse out an `[]Image` from the `values` field dynamically and a second which takes back the same `Image` type array and places the changed images in the correct spots.

internal/ecrconsumer/image_values.go
```diff
@@ -0,0 +1,114 @@
...
+func parseImage(repo string, tag string) (*Image, error) {
+	arr := strings.Split(repo, "/")
+
+	var registry string
+	var repository string
+	if len(arr) == 2 {
+		repository = arr[1]
+		registry = arr[0]
+	} else {
+		return nil, fmt.Errorf("repository field is invalid! (lacking either the full registry name or repository name)")
```
crdueck: Prefer to use `errors.New` instead of `fmt.Errorf` when there's no need for string formatting.

In this case, it would be helpful to include the malformed `repo` string, so we could use `fmt.Errorf` to add it to the error message.

internal/ecrconsumer/image_values.go
```diff
@@ -0,0 +1,114 @@
...
+			return append([]string{key}, path...)
+		}
+	}
+
+	return []string{}
+}
+
+func table(vals map[string]interface{}, path []string) map[string]interface{} {
+	tabled := vals
+	for i := range path {
```
crdueck: should use `for k,v := range kvs` style syntax here

internal/helmrelease/helmrelease.go
```diff
@@ -0,0 +1,100 @@
+package helmrelease
```
crdueck: This package can be deleted from your branch now that your code is referencing the stuff in `pkg/apis`

internal/ecrconsumer/image_values.go
```diff
@@ -0,0 +1,114 @@
+package ecrconsumer
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+
+	"ship-it/internal/helmrelease"
```
crdueck: This should reference the `v1alpha1` helmrelease pkg in `pkg/apis/...`

internal/ecrconsumer/image_values.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/google/go-github/github"
+)
+
+type GitCommands interface {
```
crdueck: This unused interface should be removed.

internal/ecrconsumer/image_values.go
```diff
@@ -0,0 +1,114 @@
...
+
+		tabled = tabled[path[i]].(map[string]interface{})
+	}
+	return tabled
+}
+
+func update(vals map[string]interface{}, img Image, path []string) map[string]interface{} {
+	imgVals := table(vals, path)
+	if imgVals == nil {
+		return nil
```
crdueck: Because we return `nil` here, when the `path` is incorrect (it's passed as a param so we have no guarantees it's actually a valid path in the helmrelease's `Spec.values`), `WithImage` would return an image with empty values. 

The preferable behaviour would be to not change the vals at all if the path doesn't match, but an ideal solution would guarantee that we can never use the wrong path in the first place. As in, we wouldn't take it by a parameter to a public func, but provide it ourselves using `getImagePath`.

internal/ecrconsumer/image_values.go
```diff
@@ -0,0 +1,123 @@
...
+func getImagePath(v reflect.Value, serviceName string) []string {
+	if v.Kind() != reflect.Map {
+		return []string{}
+	}
+
+	iter := v.MapRange()
+	for iter.Next() {
+		key := iter.Key().String()
+		if key == "image" {
+			img := v.MapIndex(iter.Key()).Interface().(map[string]interface{})
```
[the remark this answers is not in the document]
rsgc-peterson-e: I had attempted making this type change earlier but was unable to due to the following error:

```
panic: interface conversion: interface {} is map[string]interface {}, not map[string]string [recovered]
        panic: interface conversion: interface {} is map[string]interface {}, not map[string]string

goroutine 4 [running]:
testing.tRunner.func1(0xc000242100)
        /usr/local/Cellar/go/1.12.4/libexec/src/testing/testing.go:830 +0x392
panic(0x1682a20, 0xc00021b500)
        /usr/local/Cellar/go/1.12.4/libexec/src/runtime/panic.go:522 +0x1b5
ship-it/internal/ecrconsumer.getImagePath(0x166cf00, 0xc00021b1a0, 0x15, 0x171a611, 0x3, 0xc00021b1a0, 0x1d1d940, 0x1e7e008)
        /Users/ethanpeterson/Desktop/ship-it/internal/ecrconsumer/image_values.go:47 +0x62f
ship-it/internal/ecrconsumer.getImagePath(0x166cf00, 0xc00021b3e0, 0x15, 0x171a611, 0x3, 0x1cd4eac, 0x3b, 0x78077)
        /Users/ethanpeterson/Desktop/ship-it/internal/ecrconsumer/image_values.go:59 +0x215
ship-it/internal/ecrconsumer.TestGetImagePath(0xc000242100)
        /Users/ethanpeterson/Desktop/ship-it/internal/ecrconsumer/image_values_test.go:103 +0x1115
testing.tRunner(0xc000242100, 0x17487e8)
        /usr/local/Cellar/go/1.12.4/libexec/src/testing/testing.go:865 +0xc0
created by testing.(*T).Run
        /usr/local/Cellar/go/1.12.4/libexec/src/testing/testing.go:916 +0x35a
exit status 2
```

This panic happens when the code is as follows:

```
img := iter.Value().Interface().(map[string]string)
image, err := parseImage(img["repository"], img["tag"])
```

internal/helmrelease/helmrelease.go
```diff
@@ -39,3 +44,45 @@ type HelmReleaseList struct {
...
+	return h
+}
+
+func NewDecoder() runtime.Decoder {
+	factory := serializer.NewCodecFactory(runtime.NewScheme())
+
+	return factory.UniversalDeserializer()
+}
+
+func (h HelmRelease) Encode() []byte {
```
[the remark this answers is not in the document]
rsgc-peterson-e: Fixed.

internal/ecrconsumer/image_values.go
```diff
@@ -0,0 +1,123 @@
...
+	// we assume this file path until a location in miranda for custom resources is decided upon
+	resourceBytes, err := client.GetFile("master", filepath.Join("k8s/custom-resources", serviceName+".yaml"))
+	if err != nil {
+		return nil, err
+	}
+
+	rls := &helmrelease.HelmRelease{}
+	d := helmrelease.NewDecoder()
+	gvk := schema.FromAPIVersionAndKind("helmreleases.k8s.wattpad.com/v1alpha1", "HelmRelease")
+	d.Decode(resourceBytes, &gvk, rls)
```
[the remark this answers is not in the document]
rsgc-peterson-e: When running the tests again using `runtime.DecodeInto`, I observed an error with our custom resource example. The error says that a `kind` field is required in the `values` section of the yaml.

```
&errors.errorString{s:"v1alpha1.HelmRelease.Spec: v1alpha1.HelmReleaseSpec.Values: unmarshalerDecoder: Object 'Kind' is missing in '{\"autoscaler\":{\"maxPods\":50,\"minPods\":30,\"targetCPUUtilizationPercent\":60},\"containerPort\":80,\"cronjob\":{\"closeoutAppLabel\":\"loki-closeout\",\"image\":{\"repository\":\"723255503624.dkr.ecr.us-east-1.amazonaws.com/kube-tools\",\"tag\":\"deda27d\"},\"schedule\":\"0 0 * * *\"},\"image\":{\"repository\":\"723255503624.dkr.ecr.us-east-1.amazonaws.com/loki\",\"tag\":\"cc064f8a3d3fa0fe938e95d961ad0278770fa5d2\"},\"microservice\":{\"nameOverride\":\"loki\"},\"nodePort\":31828,\"resources\":{\"limits\":{\"cpu\":\"500m\",\"memory\":\"256Mi\"},\"requests\":{\"cpu\":\"500m\",\"memory\":\"128Mi\"}},\"securityContext\":{\"privileged\":true},\"serviceAccountName\":\"loki\",\"servicePort\":80}', error found in #10 byte of ...|ePort\":80}}}|..., bigger context ...|rue},\"serviceAccountName\":\"loki\",\"servicePort\":80}}}|..."}
```

I fixed the error by adding the field in the test yaml like so:

```
apiVersion: helmreleases.k8s.wattpad.com/v1alpha1
kind: HelmRelease
metadata:
  name: example-microservice
spec:
  chart:
    repository: wattpad.s3.amazonaws.com/helm-charts
    path: microservice
    revision: HEAD
  releaseName: example-release
  values:
    kind: HelmRelease
    autoscaler:
      maxPods: 50
      minPods: 30
      targetCPUUtilizationPercent: 60
    containerPort: 80
    cronjob:
      closeoutAppLabel: loki-closeout
      image:
        repository: 723255503624.dkr.ecr.us-east-1.amazonaws.com/kube-tools
        tag: deda27d
      schedule: 0 0 * * *
    image:
      repository: 723255503624.dkr.ecr.us-east-1.amazonaws.com/loki
      tag: cc064f8a3d3fa0fe938e95d961ad0278770fa5d2
    microservice:
      nameOverride: loki
    nodePort: 31828
    resources:
      limits:
        cpu: 500m
        memory: 256Mi
      requests:
        cpu: 500m
        memory: 128Mi
    securityContext:
      privileged: true
    serviceAccountName: loki
    servicePort: 80
```

internal/ecrconsumer/image_values.go
```diff
@@ -0,0 +1,156 @@
...
+func update(vals map[string]interface{}, img Image) map[string]interface{} {
+	path := getImagePath(reflect.ValueOf(vals), img.Repository)
+	if len(path) == 0 {
+		return nil
+	}
+	imgVals := table(vals, path)
+	if imgVals == nil {
+		return nil
+	}
+	imgVals["repository"] = img.Registry + "/" + img.Repository
```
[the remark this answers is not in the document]
rsgc-peterson-e: Perhaps `RepoURI()` could be a good name? For now I have called the method `Untagged()`.

internal/ecrconsumer/image_values_test.go
```diff
@@ -0,0 +1,401 @@
...
+	}
+	for i, tc := range tests {
+		output := v1alpha1.HelmValues(make(map[string]interface{}))
+		tc.original.DeepCopyInto(&output)
+		assert.Exactly(t, tc.expectedCopy, tc.transformer(output), "copy was not mutated. test case: %d", i)
+		assert.Exactly(t, tc.expectedOriginal, tc.original, "original was mutated. test case: %d", i)
+	}
+}
+
+func TestFullCycle(t *testing.T) {
```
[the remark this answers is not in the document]
rsgc-peterson-e: I renamed the function to `TestSerializeRoundTrip`. However, running the test in the `v1alpha1` package creates a circular import due to the way the project is currently configured.

pkg/apis/k8s.wattpad.com/v1alpha1/decoder_test.go
```diff
@@ -0,0 +1,118 @@
...
+        cpu: 500m
+        memory: 128Mi
+    securityContext:
+      privileged: true
+    serviceAccountName: loki
+    servicePort: 80
+status: {}
+`
+
+func LoadRelease(fileData []byte) (*HelmRelease, error) {
```
crdueck: Let's make this test helper private.

internal/ecrconsumer/image_values.go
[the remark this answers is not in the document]
crdueck: I think `URI` is a good name, that's what ECR console calls this untagged registry+repo string.

internal/ecrconsumer/image_values_test.go
```diff
@@ -0,0 +1,325 @@
...
+			Values: map[string]interface{}{
+				"image": map[string]interface{}{
+					"repository": "bar/foo",
+					"tag":        "baz",
+				},
+			},
+		},
+	}
+
+	outputRls := WithImage(expectedImg, rls)
```
crdueck: Couple thoughts on this test case. It's an important one, since the goal of this PR was to implement `WithImage`.

The overall behaviour we want to test is that `WithImage` finds and sets the desired image tag if there's a matching repo, and does nothing if there's no match. So we should have a test for both scenarios.

Second, we should avoid re-implementing the business logic we're trying to test, in the test case. `WithImage` calls `update` -> `getImagePath` & `table`, which are called again in the test case.

We know what our test data looks like, so we should take advantage of that:

```
outputRls := WithImage(expectedImg, rls)
outputImg := outputRls.Spec.Values["image"].(map[string]interface{})
assert.Equal(t, expectedImg.Tag, outputImg["tag"].(string)
```

For the "no image repo matches desired image" test case, we can assert that the original and modified HelmRelease are the same. Something like this should work:

```
// same setup as original TestWithImage
outputRls := WithImage(expectedImg, rls)
assert.Exactly(t, rls, outputRls)
```

internal/ecrconsumer/image_values_test.go
```diff
@@ -302,12 +302,31 @@ func TestWithImage(t *testing.T) {
...
 	}
 
-	imgVals := table(outputRls.Spec.Values, path)
+	outputImg := outputRls.Spec.Values["image"].(map[string]interface{})
+	assert.Equal(t, expectedImg.Tag, outputImg["tag"].(string))
 
-	outputImage, err := parseImage(imgVals["repository"].(string), imgVals["tag"].(string))
-	assert.NoError(t, err)
+	// Test No Matching Image Case
+	rls = v1alpha1.HelmRelease{
```
crdueck: The match/no match cases should be done in separate sub-tests, so it's obvious what actually caused `TestWithImage` to fail. Usually I'd split them into separate test funcs, but since they share some test data I think it makes sense to group them together.

We can create a sub-test scope with 
```
t.Run("descriptive name", func(t *testing.T) { 
    // sub test
})
```

internal/ecrconsumer/image_values_test.go
```diff
@@ -302,12 +302,31 @@ func TestWithImage(t *testing.T) {
...
+			Name: "example-microservice",
+		},
+		Spec: v1alpha1.HelmReleaseSpec{
+			ReleaseName: "example-release",
+			Chart: v1alpha1.ChartSpec{
+				Repository: "wattpad.s3.amazonaws.com/helm-charts",
+				Path:       "microservice",
+				Revision:   "HEAD",
+			},
+			Values: map[string]interface{}{},
```
crdueck: The `Values` shouldn't be empty, it defeats the point of the test case which is that non-matching image values shouldn't be changed. We don't really test that if there's no image values at all. If we want a 3rd case for no values, that could be a separate test but I don't think it's necessary.

Also I would avoid duplicating the big data (the HelmRelease struct) and rather duplicate the small data (desired Image) instead. We could use the same release struct in both cases and have 2 images, one of which wouldn't match the release.

internal/ecrconsumer/image_values_test.go
```diff
@@ -0,0 +1,344 @@
...
+					"repository": "bar/foo",
+					"tag":        "baz",
+				},
+			},
+		},
+	}
+
+	outputRls := WithImage(expectedImg, rls)
+
+	path := getImagePath(outputRls.Spec.Values, "foo")
```
crdueck: `path` var is now unused and should be removed

PR #29, opened 2019-06-21
VELO-1420: Add HelmRelease CustomResourceDefinition

crdueck:


deploy/v1alpha1_helmrelease_crd.yaml
```diff
@@ -0,0 +1,38 @@
...
+  version: v1alpha1
+  names:
+    kind: HelmRelease
+    listKind: HelmReleaseList
+    plural: helmreleases
+    singular: helmrelease
+    shortNames:
+      - rls
+  subresources:
+    status: {}
```
crdueck: This subresource will allow us to put the helm release state into the CRs for the API service to query.
https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#status-subresource

crdueck:
I agree that the CRs should contain metadata about the release, like the owner, monitoring etc. Since these details aren't required to actually do the deployment, I was thinking they could be annotations on the CR instead of being part of the CRD.

When the UI queries the CRs, it has access to their annotations through the [ObjectMeta](https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta.GetAnnotations)

crdueck:
@cfmtam I added some example annotations based on the fields we want to include in the `/releases` response. I realized that it might be messy to provide the DD monitor URLs. Since the `annotations` section is a `map[string]string`, we can't do something like 

```
helmreleases.k8s.wattpad.com/monitors: [url1, url2, url3]
```

but maybe we could do

```
helmreleases.k8s.wattpad.com/monitors/monitor1: url1
helmreleases.k8s.wattpad.com/monitors/monitor2: url2
helmreleases.k8s.wattpad.com/monitors/monitor3: url3
```

or ... csv string

```
helmreleases.k8s.wattpad.com/monitors: "url1, url2, url3"
```

I'm not sure which one is least bad. Any other alternatives? I do like the separation of required values in the spec body and optional metadata in the annotations, but it's a bit awkward here.

crdueck:
I'm gonna leave monitors out of the example for now. We can revisit once the UI is querying these CRs.

PR #30, opened 2019-06-24
Project re-organization

crdueck:
This changeset breaks up the API server and ecrconsumer. Previously they were part of the same binary/container, but we want to avoid coupling between the core deployment systems and the UI service.

The `internal` pkg is also re-organized so that all the api-specific integration & model pkgs are children of the `api` pkg. The pkg structure now looks like:
```
── internal
   ├── api
   │   ├── config
   │   ├── integrations
   │   │   ├── github
   │   │   └── k8s
   │   ├── middleware
   │   ├── models
   │   └── service
   └── ecrconsumer
       └── config
```


cmd/ship-it-ecrconsumer/main.go
```diff
@@ -0,0 +1,58 @@
...
+	"ship-it/internal/ecrconsumer"
+	"ship-it/internal/ecrconsumer/config"
+
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/sqs"
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/metrics/dogstatsd"
+)
+
+func main() {
```
crdueck: Lots of duplication between `main`s, though it's all boilerplate service init. We could consider abstracting this later.

Makefile
```diff
@@ -1,20 +1,21 @@
...
-IMAGE := $(PROJECT_NAME):$(VERSION)
-LATEST_IMAGE := $(PROJECT_NAME):latest
+env-target:
+ifndef TARGET
+	    $(error TARGET is undefined)
+endif
 
-build:
-	docker build -t $(IMAGE) .
+build: env-target
```
crdueck: Is there a better way to manage building separate targets? This works, but it becomes awkward to have a single "build all the things" target for travis.

jrhouston: Yeah I don't think there's much better we can do with Makefiles, I think this is OK for now.

PR #33, opened 2019-06-28
VELO-1457: lister integration

crdueck:
This PR updates the API's k8s integration to use the generated HelmReleases informer instead of querying config maps.

---
This is still untested, but the code changes are pretty simple. I should test this out in minikube.

### DIFF
--- a/internal/api/integrations/k8s/k8s.go
+++ b/internal/api/integrations/k8s/k8s.go
@@ -10,7 +10,7 @@ import (
 )
 
 type K8sClient struct {
-	helmreleases v1alpha1.HelmreleasesV1alpha1Interface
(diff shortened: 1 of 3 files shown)


internal/api/integrations/k8s/k8s.go
```diff
@@ -1,47 +1,63 @@
...
+	exit := make(chan struct{})
 
-type K8sClient struct {
-	core kv1.CoreV1Interface
+	go func() {
+		<-ctx.Done()
+		exit <- struct{}{}
+	}()
+
+	factory := informers.NewSharedInformerFactory(client, 30*time.Second)
```
jrhouston: Can we make this a config option instead of having this magic number here.

PR #42, opened 2019-07-05
Load all available fields to populate /releases endpoint for UI demo

rsgc-peterson-e:
VELO-1465


internal/api/integrations/k8s/k8s.go
```diff
@@ -42,6 +43,10 @@ func New(ctx context.Context, resync time.Duration) (*K8sClient, error) {
 	}, nil
 }
 
+func getKey(k string) string {
+	return "helmreleases." + k8swattpadcom.GroupName + "/" + k
```
crdueck:
```suggestion
	return v1alpha1.Resource("helmreleases").String() + "/" + k
```

PR #43, opened 2019-07-08
Commit CRs to GitHub VELO-1454

rsgc-peterson-e:


internal/syncd/integrations/ecr/reconciler.go
```diff
@@ -2,27 +2,55 @@ package ecr
...
-	GetContents(ctx context.Context, org, repo, path string, opts *github.RepositoryContentGetOptions) (*github.RepositoryContent, []*github.RepositoryContent, *github.Response, error)
-	UpdateFile(ctx context.Context, owner, repo, path string, opt *github.RepositoryContentFileOptions) (*github.RepositoryContentResponse, *github.Response, error)
+	UpdateFile(msg string, branch string, path string, fileContent []byte) (*github.RepositoryContentResponse, error)
+	GetFile(branch string, path string) ([]byte, error)
 }
 
 type ImageReconciler struct {
-	// TODO
+	Org          string
+	ResourcePath string
```
crdueck: This approach will only work for a single resource on a single branch. We need to be able to make changes to multiple HelmRelease files (probably for the same branch, `master`), so the path to the CR will probably need to be configured from env.

The branch can be fixed, but it should be configurable from env and we can default it to `master`

internal/syncd/integrations/ecr/reconciler.go
```diff
@@ -2,27 +2,55 @@ package ecr
...
+	rls, err := v1alpha1.LoadRelease(resourceBytes)
+	if err != nil {
+		return err
+	}
+
+	updatedRls := WithImage(*image, *rls)
+
+	updatedBytes, err := yaml.Marshal(updatedRls)
+	if err != nil {
+		return nil
```
crdueck: `return err`?

internal/syncd/integrations/ecr/reconciler.go
```diff
@@ -2,27 +2,55 @@ package ecr
...
+		ResourcePath: valPath,
+		RepoService:  r,
 	}
 }
 
 func (r *ImageReconciler) Reconcile(ctx context.Context, image *internal.Image) error {
-	return nil
+	resourceBytes, err := r.RepoService.GetFile(r.Branch, r.ResourcePath)
+	if err != nil {
+		return err
```
crdueck: Let's add context to all these `return err` with `errors.Wrap`. 

My rule of thumb is that if I'm returning an error from an internal package (code we control), I usually don't add context because I assume the internal code already adds that context. If I'm calling 3rd party code, like this github `RepositoriesService`, then I usually recommend to add context to the bare error.

internal/syncd/integrations/ecr/listener.go
```diff
@@ -18,6 +23,13 @@ type ImageListener struct {
 	timer   metrics.Histogram
 }
 
+type SQSMessage struct {
```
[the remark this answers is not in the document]
rsgc-peterson-e: Yes, I thought excluding the struct tags would cause an issue as well. However, I wrote a test case for the parse message function, which unmarshals into the struct and it passes without the json tags.
```
func TestParseMessage(t *testing.T) {
	inputJSON := `
{
	"eventTime": "2019-07-11T14:19:59Z", 
	"repositoryName": "monolith-php", 
	"tag": "78bc9ccf64eb838c6a0e0492ded722274925e2bd",
	"registryId": "723255503624"
}
`

	deployTime, err := time.Parse(time.RFC3339, "2019-07-11T14:19:59Z")
	assert.NoError(t, err)

	expectedMessage := SQSMessage{
		EventTime:      deployTime,
		RepositoryName: "monolith-php",
		Tag:            "78bc9ccf64eb838c6a0e0492ded722274925e2bd",
		RegistryId:     "723255503624",
	}

	inputMessage, err := parseMsg(inputJSON)
	assert.NoError(t, err)
	assert.Exactly(t, expectedMessage, *inputMessage)
}
```

Note that the `SQSMessage` type is renamed to `ecrPushEvent`

internal/syncd/integrations/ecr/reconciler.go
rsgc-peterson-e: I have added some ECR related environment variables to the `syncd` config struct. The release branch is an environment variable with a default value of `master`

internal/syncd/integrations/ecr/listener.go
```diff
@@ -40,6 +52,44 @@ func (l *ImageListener) Listen(ctx context.Context, r syncd.ImageReconciler) err
 	return sqsconsumer.NewConsumer(l.service, stack).Run(ctx)
 }
 
+func parseMsg(msg string) (*SQSMessage, error) {
```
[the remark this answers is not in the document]
rsgc-peterson-e: Done.

internal/syncd/config/config.go
```diff
@@ -20,6 +20,8 @@ type Config struct {
 	Namespace          string `split_words:"true" required:"true"`
 	ReleaseName        string `split_words:"true" required:"true"`
 	HelmTimeoutSeconds int64  `split_words:"true" default:"10"`
+	RegistryChartPath  string `required:"true"`
```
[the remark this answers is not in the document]
rsgc-peterson-e: Yes it does need the flag so envconfig looks for `REGISTRY_CHART_PATH`. I have also made this correction for the release branch environment variable.

internal/syncd/integrations/ecr/reconciler.go
```diff
@@ -24,23 +25,25 @@ type ImageReconciler struct {
...
-		ResourcePath: valPath,
+		ResourcePath: prefix,
+		Branch:       branch,
 		RepoService:  r,
 	}
 }
 
 func (r *ImageReconciler) Reconcile(ctx context.Context, image *internal.Image) error {
-	resourceBytes, err := r.RepoService.GetFile(r.Branch, r.ResourcePath)
+	path := filepath.Join(r.ResourcePath, image.Repository+".yaml")
```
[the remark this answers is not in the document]
rsgc-peterson-e: Indeed, I see this PR is also blocked by the difference between release and repository names. Once we have a better idea of how to go about fixing this issue, I will fix this PR and the Travis PR accordingly.

internal/syncd/integrations/ecr/reconciler.go
```diff
@@ -2,27 +2,57 @@ package ecr
...
+		Org:               org,
+		RegistryChartPath: prefix,
+		Branch:            branch,
+		RepoService:       r,
 	}
 }
 
 func (r *ImageReconciler) Reconcile(ctx context.Context, image *internal.Image) error {
-	return nil
+	path := filepath.Join(r.RegistryChartPath, image.Repository+".yaml")
```
crdueck: Here is where we would use the upcoming image repositories informer to build this file name. For each release affected by the updated image, we need to update each of the release `.yaml` files.

internal/syncd/integrations/ecr/reconciler_test.go
```diff
@@ -0,0 +1,271 @@
...
+type MockIndexerService struct {
+	mock.Mock
+}
+
+func (m *MockIndexerService) Lookup(repo string) ([]types.NamespacedName, error) {
+	args := m.Called(repo)
+	return args.Get(0).([]types.NamespacedName), args.Error(1)
+}
+
+func TestReconcile(t *testing.T) {
```
rsgc-peterson-e: There is a lot of code repetition throughout these test cases. I tried using named sub-tests to avoid repeatedly mocking out the same dependencies. However, it seems when I call `mock.On()` twice in the same function, only the first call takes effect instead of the second one taking over in the following named sub test. As a result, I must place the cases in different functions entirely. Another work around for this would be to declare the common dependencies as global variables but I am unsure if this is considered a good practice.

internal/syncd/integrations/ecr/git.go
```diff
@@ -0,0 +1,57 @@
...
+	client := GitHub{
+		client:       github.NewClient(oauth2.NewClient(ctx, tokenSource)),
+		Organization: org,
+		Repository:   repo,
+	}
+
+	return client
+}
+
+func (c GitHub) UpdateFile(msg string, branch string, path string, fileContent []byte) (*github.RepositoryContentResponse, error) {
```
rsgc-peterson-e: Should test cases be written for these functions? They do not have much logic to test, however, the code coverage would be greatly improved on this PR by adding them.

internal/syncd/integrations/ecr/reconciler_test.go
```diff
@@ -0,0 +1,271 @@
...
+		Tag:        "78bc9ccf64eb838c6a0e0492ded722274925e2bd",
+	}
+
+	mockRepoService.On("GetFile", fakeReconciler.Branch, filepath.Join(fakeReconciler.RegistryChartPath, inputImage.Repository+".yaml")).Return("some malformed yaml", error(nil))
+	err := updateResource(fakeReconciler, inputImage, "bar")
+	assert.Error(t, err)
+	mockRepoService.AssertExpectations(t)
+}
+
+func TestNewReconciler(t *testing.T) {
```
crdueck: I don't think this test adds much value. If we want code coverage for the `NewReconciler` func, we should be calling it in the other test cases instead of constructing `ImageReconciler` with composite-literal syntax.

internal/syncd/integrations/ecr/listener_test.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	fakeListener := &ImageListener{
+		logger:  log.NewNopLogger(),
+		service: mockSQSService,
+		timer:   fakeHist,
+	}
+
+	mockRepoService := new(MockRepoService)
+	fakeIndexService := new(MockIndexerService)
+	fakeReconciler := NewReconciler("Wattpad", "custom-resources/path", "foo", mockRepoService, fakeIndexService)
```
crdueck: As a direct dependency of the `ImageListener`, the `ImageReconciler` should be mocked out instead of mocking the transitive dependencies on `RepoService` and `IndexerService`

internal/syncd/integrations/ecr/listener_test.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	for _, test := range tests {
+		assert.Equal(t, test.expected, validImageTagRegex.MatchString(test.inputString))
+	}
+}
+
+func TestECRHandler(t *testing.T) {
+	mockSQSClient := new(MockSQS)
+	fakeURL := "www.wattpad.com"
+	var fakeHist metrics.Histogram
```
crdueck: use [`metrics/discard`](https://godoc.org/github.com/go-kit/kit/metrics/discard) for nop metric implementations.

internal/syncd/integrations/ecr/reconciler_test.go
rsgc-peterson-e: Removed it and called `NewReconciler` in the other test cases.

internal/syncd/integrations/ecr/reconciler.go
```diff
@@ -2,27 +2,78 @@ package ecr
...
+	}
+
+	rls := make(map[string]interface{})
+
+	err = yaml.Unmarshal([]byte(resourceStr), rls)
+	if err != nil {
+		return errors.Wrapf(err, "failed to parse the YAML file")
+	}
+
+	updatedRls := internal.WithImage(*image, rls)
```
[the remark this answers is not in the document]
rsgc-peterson-e: You are correct, I set the parameter to be a `map[string]interface{}` to get around the `runtime.RawExtension` type issues we were having earlier. Since the function does not depend on the `HelmRelease` type, the function is greatly simplified by using a generic `map[string]interface{}` and the standard `yaml` package.

internal/image.go
```diff
@@ -27,3 +27,108 @@ func ParseImage(repo string, tag string) (*Image, error) {
...
+	case []interface{}:
+		return cleanUpInterfaceArray(v)
+	case map[interface{}]interface{}:
+		return cleanUpInterfaceMap(v)
+	default:
+		return v
+	}
+}
+
+func DeepCopyMap(in map[string]interface{}) map[string]interface{} {
```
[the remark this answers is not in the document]
rsgc-peterson-e: My mistake, I have replaced the function with a recursive solution.

internal/syncd/integrations/ecr/git.go
```diff
@@ -0,0 +1,57 @@
...
+		Branch:  github.String(branch),
+	}
+
+	resp, _, err := c.client.Repositories.UpdateFile(ctx, c.Organization, c.Repository, path, options)
+
+	return resp, err
+}
+
+func (c GitHub) GetFile(branch string, path string) (string, error) {
+	contents, _, _, err := c.client.Repositories.GetContents(context.Background(), c.Organization, c.Repository, path, &github.RepositoryContentGetOptions{Ref: "refs/heads/" + branch})
```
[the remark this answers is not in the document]
rsgc-peterson-e: Fixed, I have also made both functions private to the package.

internal/image_test.go
```diff
@@ -37,3 +37,322 @@ func TestParseImage(t *testing.T) {
...
+			},
+		},
+	}
+	for _, test := range tests {
+		update(test.inputMap, test.newImage)
+		assert.Equal(t, test.expectedMap, test.inputMap)
+	}
+}
+
+func TestWithImage(t *testing.T) {
```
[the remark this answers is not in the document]
rsgc-peterson-e: Agreed, I had copied the old tests from `ecrconsumer` before writing the tests for `WithImage`. The `WithImage` tests cover the same behaviour so I have removed the `UpdateImage` tests. Code coverage is also unaffected by this change.

internal/syncd/integrations/ecr/reconciler.go
```diff
@@ -2,27 +2,78 @@ package ecr
...
+		RegistryChartPath: prefix,
+		Branch:            branch,
+		RepoService:       r,
+		IndexService:      i,
 	}
 }
 
+func updateResource(r *ImageReconciler, image *internal.Image, name string) error {
+	path := filepath.Join(r.RegistryChartPath, name+".yaml")
+	resourceStr, err := r.RepoService.GetFile(r.Branch, path)
```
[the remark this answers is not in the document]
rsgc-peterson-e: Moved the branch field to the Github type.

internal/syncd/integrations/ecr/reconciler.go
```diff
@@ -2,27 +2,46 @@ package ecr
...
+	RegistryChartPath string
+	RepoService       RepositoriesService
+	IndexService      IndexerService
 }
 
-func NewReconciler(org string, r RepositoriesService) *ImageReconciler {
+func NewReconciler(prefix string, r RepositoriesService, i IndexerService) *ImageReconciler {
 	return &ImageReconciler{
-		// TODO
+		RegistryChartPath: prefix,
```
crdueck: The `RegistryChartPath` field exists only to be passed to the `RepoService.UpdateAndReplace` method call. This suggests that the field should belong to the `RepoService` instead.

I think the `RepositoriesService` interface name should be updated as well, since the old one was a reference to go-github's [RepositoriesService](https://godoc.org/github.com/google/go-github/github#RepositoriesService). Something like `HelmReleaseEditor` or `HelmReleaseUpdater`  would make sense to me

internal/syncd/integrations/ecr/reconciler_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+func (m *MockIndexerService) Lookup(repo string) ([]types.NamespacedName, error) {
+	args := m.Called(repo)
+	return args.Get(0).([]types.NamespacedName), args.Error(1)
+}
+
+func TestReconcileLookupFailure(t *testing.T) {
+	mockRepoService := new(MockRepoService)
+	mockIndexService := new(MockIndexerService)
+	fakeReconciler := NewReconciler("foo/bar/resources", mockRepoService, mockIndexService)
```
crdueck: This isn't really a fake, it's a real reconciler that uses mock dependencies.

I've used the term `fake` in other places like [here](https://github.com/Wattpad/ship-it/pull/68/files#diff-83d733c3663e96ac502b181ebf9d47d4R34) to distinguish between a true mock and a fake implementation. The `fakeCache` there is a real implementation of a cache, no mocks, but it has limited functionality and it's purely in-memory to support unit testing.

I don't think we should call the thing we're testing "fake". It's very suggestive and makes it less clear what's actually being tested vs. the dependencies.

internal/syncd/integrations/ecr/sqsmock.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"github.com/aws/aws-sdk-go/service/sqs"
+	"github.com/stretchr/testify/mock"
+)
+
+type MockSQS struct {
+	mock.Mock
+}
+
+type SQSAPI interface {
```
crdueck: Is this interface definition needed? Elsewhere the `sqsconsumer.SQSAPI` interface is used instead.

internal/syncd/integrations/ecr/reconciler_test.go
rsgc-peterson-e: Agreed, I will rename it to `mockReconciler`

internal/syncd/integrations/ecr/reconciler.go
rsgc-peterson-e: Renamed to `HelmReleaseEditor`

internal/syncd/integrations/ecr/reconciler_test.go
```diff
@@ -69,7 +69,7 @@ func (m *MockIndexerService) Lookup(repo string) ([]types.NamespacedName, error)
 func TestReconcileLookupFailure(t *testing.T) {
 	mockRepoService := new(MockRepoService)
 	mockIndexService := new(MockIndexerService)
-	fakeReconciler := NewReconciler("foo/bar/resources", mockRepoService, mockIndexService)
+	mockReconciler := NewReconciler(mockRepoService, mockIndexService)
```
crdueck: This reconciler isn't a mock

internal/syncd/integrations/ecr/reconciler_test.go
```diff
@@ -117,7 +117,7 @@ func TestReconcilerSuccess(t *testing.T) {
 	mockRepoService := new(MockRepoService)
 	mockIndexService := new(MockIndexerService)
 
-	fakeReconciler := NewReconciler("foo/bar/resources", mockRepoService, mockIndexService)
+	mockReconciler := NewReconciler(mockRepoService, mockIndexService)
```
crdueck: This reconciler isn't a mock

internal/syncd/integrations/ecr/reconciler.go
```diff
@@ -2,27 +2,42 @@ package ecr
...
+	UpdateAndReplace(ctx context.Context, releaseName string, image *internal.Image) error
+}
+
+type IndexerService interface {
+	Lookup(repo string) ([]types.NamespacedName, error)
 }
 
 type ImageReconciler struct {
-	// TODO
+	RepoService  HelmReleaseEditor
```
crdueck: `RepoService` field should be renamed

internal/syncd/integrations/ecr/reconciler_test.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	mockIndexService.On("Lookup", "bar").Return([]types.NamespacedName{
+		{
+			Namespace: "default",
+			Name:      "bar",
+		},
+	}, error(nil))
+
+	mockRepoService.On("UpdateAndReplace", mock.Anything, mock.Anything, inputImage).Return(fmt.Errorf("some update image error"))
```
crdueck: We should expect the second argument to `UpdateAndReplace` will be the release name `"bar"`

internal/syncd/integrations/ecr/reconciler_test.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	mockIndexService.On("Lookup", "bar").Return([]types.NamespacedName{
+		{
+			Namespace: "default",
+			Name:      "bar",
+		},
+	}, error(nil))
+
+	mockRepoService.On("UpdateAndReplace", mock.Anything, mock.Anything, inputImage).Return(error(nil))
```
crdueck: We should expect the second argument to UpdateAndReplace will be the release name "bar"

crdueck: Is there a reason to coerce `error(nil)` instead of just `nil` with an inferred type?

internal/syncd/integrations/ecr/reconciler.go
```diff
@@ -2,27 +2,42 @@ package ecr
...
 }
 
 func (r *ImageReconciler) Reconcile(ctx context.Context, image *internal.Image) error {
+	releases, err := r.IndexService.Lookup(image.Repository)
+	if err != nil {
+		return errors.Wrapf(err, "failed to obtain the releases corresponding to the repository: %s", image.Repository)
+	}
+	for _, release := range releases {
+		err := r.RepoService.UpdateAndReplace(ctx, release.Name, image)
+		if err != nil {
```
crdueck: We should think more about the behaviour here. We would probably like to continue trying to update the rest of the releases and aggregate errors as we go (we could use the `syncd.multiError` type).

Let's use this logic for now and come back to it.

rsgc-peterson-e: For sure! Happy to open another PR adjusting this, I agree it is not a good behaviour for the entire update to fail as soon as an error is thrown.

internal/syncd/integrations/ecr/reconciler_test.go
rsgc-peterson-e: No, seems to work without the casting. Fixed it.

internal/syncd/integrations/ecr/reconciler.go
rsgc-peterson-e: Renamed to `Editor`

PR #44, opened 2019-07-08
VELO-1452: github event consumer

crdueck:


internal/gitsync/github/handler_test.go
```diff
@@ -0,0 +1,81 @@
+package github
```
crdueck: These test cases are honestly pretty bad, I basically wrote them just for the coverage. There's not really any higher-level properties to test here, just checking that it handles errors sanely and calls methods on its dependencies as expected.

internal/gitsync/github/release_syncer_test.go
```diff
@@ -0,0 +1,71 @@
+package github
```
crdueck: these tests are questionable as well, maybe a bit better than the handler tests.

Ideally the test would be show that after calling `UpdateOrInstallFromChart`, we can observe that the release "exists" regardless of whether it was installed from nothing, or upgraded from an existing release. That'd be a much better behavioural test that what i'm doing here, checking that the right helm client methods are called.

Using a mock `helmClient` to test that the release exists seems pointless, I'm basically testing the mock at that point. So I'm not sure what a better way to write these tests would be.

PR #48, opened 2019-07-11
Add travis build URL to /releases endpoint VELO-1473

rsgc-peterson-e:


internal/api/service/service.go
```diff
@@ -9,18 +9,28 @@ import (
...
 }
 
+type GitCommands interface {
+	GetTravisCIBuildURLForRef(ctx context.Context, repo string, ref string) (string, error)
+}
+
 type Service struct {
-	lister ReleaseLister
+	lister     ReleaseLister
+	client     GitCommands
```
crdueck: `client` & `GitCommands` leak implementation details of the interface unnecessarily. An interface should provide a black box abstraction. When we use it, we needn't care that it's implemented by a `github.Client` and we don't even need to know it's related to git/github at all.

As far as we (the `Service` consuming the interface) need to know, the interface integrates with the Travis Checks API. So, I'd suggest naming the interface `TravisChecker` (go interfaces are [conventionally](https://golang.org/doc/effective_go.html#interface-names) named this way), and the struct field `travis`.

internal/api/integrations/k8s/k8s.go
```diff
@@ -58,11 +64,14 @@ func (k *K8sClient) ListAll(namespace string) ([]models.Release, error) {
...
 		annotations := r.GetAnnotations()
 		autoDeploy, err := strconv.ParseBool(annotations[annotationFor("autodeploy")])
-		if err != nil {
-			return nil, err
-		}
+
+		k.logger.Log(err)
+
+		codeURL := annotations[annotationFor("code")]
+		serviceName := r.GetName()
```
crdueck: I think `releaseName` would be more suitable, since it's `Release.GetName()`

internal/api/integrations/k8s/k8s.go
```diff
@@ -76,16 +85,44 @@ func (k *K8sClient) ListAll(namespace string) ([]models.Release, error) {
...
+
+func findVCSRepo(url string) string {
+	arr := strings.Split(url, "/")
+	if len(arr) > 4 {
+		return arr[4]
+	}
+	return ""
+}
+
+func GetImageForRepo(repo string, vals map[string]interface{}, logger log.Logger) internal.Image {
```
crdueck: Instead of taking a logger argument, this func should return an `error` and let the caller log it. I'd also suggest returning `(*Image, error)` so we can return `nil` in the error case.

internal/api/service/service.go
```diff
@@ -56,9 +55,10 @@ func (s *Service) getReleaseStatus(_ *models.Release) release.Status_Code {
 	return release.Status_PENDING_INSTALL
 }
 
-func (s *Service) getTravisURL(ctx context.Context) string {
-	url, err := s.client.GetTravisCIBuildURLForRef(ctx, s.repository, s.ref)
+func (s *Service) getTravisURL(ctx context.Context, r models.Release) string {
```
crdueck: I think it makes sense to take a `*models.Release`, since the Release struct is much larger than a pointer, and we avoid awkwardly dereferencing the reference we just created in the for-loop of `ListReleases`

internal/api/service/service_test.go
```diff
@@ -30,13 +32,28 @@ func newMockK8sClient(name string, time time.Time) *mockK8sClient {
...
+func TestListReleasesPopulatesFromCustomResource(t *testing.T) {
+	name := "mock-resource"
 	currentTime := time.Now()
 
-	mock := newMockK8sClient(name, currentTime)
-	mockGit := github.New(context.Background(), "Wattpad", "fake")
-	svc := New(mock, mockGit, "miranda", "master")
+	mockK8s := newMockK8sClient(name, currentTime)
+	mockGit := newMockGitClient("wattpad")
+	svc := New(mockK8s, mockGit, log.NewJSONLogger(log.NewSyncWriter(os.Stdout)))
```
crdueck: Use `log.NewNopLogger()` in test cases

internal/api/integrations/k8s/k8s.go
```diff
@@ -58,11 +64,14 @@ func (k *K8sClient) ListAll(namespace string) ([]models.Release, error) {
 	for _, r := range releaseList {
 		annotations := r.GetAnnotations()
 		autoDeploy, err := strconv.ParseBool(annotations[annotationFor("autodeploy")])
-		if err != nil {
-			return nil, err
-		}
+
+		k.logger.Log(err)
```
crdueck: This error needs more context. Seeing `unable to parse "ture" as bool` (or something like that) in the log output of this service won't be useful or actionable. We should use [`errors.Wrap`](https://godoc.org/github.com/pkg/errors#Wrapf) to provide more detail, like what caused the parse failure (parsing autodeploy annotation) and which release contains the invalid annotation

We should also only log the error if it's `!= nil`. The logger also needs 2 arguments, a key-value pair. For simple error handling like this, we usually use `Log("error", err)`

internal/api/integrations/k8s/k8s.go
```diff
@@ -76,16 +85,44 @@ func (k *K8sClient) ListAll(namespace string) ([]models.Release, error) {
...
+				},
 			},
 		})
 	}
 
 	return releases, nil
 }
+
+func findVCSRepo(url string) string {
+	arr := strings.Split(url, "/")
```
crdueck: I think we should implement this func by [parsing](https://golang.org/pkg/net/url/#Parse) the URL (to validate it's a real URL). Then getting the repo split from the `URL.Path`

internal/api/integrations/k8s/k8s.go
```diff
@@ -76,16 +85,44 @@ func (k *K8sClient) ListAll(namespace string) ([]models.Release, error) {
...
+func GetImageForRepo(repo string, vals map[string]interface{}, logger log.Logger) internal.Image {
+	arr := internal.GetImagePath(vals, repo)
+	if len(arr) == 0 {
+		logger.Log("No Image Found for repository")
+		return internal.Image{}
+	}
+	imgVals := internal.Table(vals, arr)
+	img, err := internal.ParseImage(imgVals["repository"].(string), imgVals["tag"].(string))
+	if err != nil {
+		logger.Log("Unable to parse valid image")
```
crdueck: This message is confusing, we couldn't parse a _valid_ image? This log message should include the error message as well.

internal/api/integrations/k8s/k8s_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"testing"
+
+	"ship-it/internal"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAnnotationFor(t *testing.T) {
```
crdueck: Let's write these table-driven tests using named sub-tests, like so: https://github.com/Wattpad/ship-it/blob/master/internal/image_test.go#L9

I personally prefer to define named type, rather than an inline anonymous struct type, but it's fairly subjective.

internal/api/service/service_test.go
```diff
@@ -29,12 +31,28 @@ func newMockK8sClient(name string, time time.Time) *mockK8sClient {
...
+}
+
+func newMockGitClient(org string) *mockGitClient {
+	return &mockGitClient{
+		Org: org,
+	}
+}
+
+func (m *mockGitClient) GetTravisCIBuildURLForRef(ctx context.Context, repo string, ref string) (string, error) {
+	return "www.travisci.com", nil
```
crdueck: This isn't really a test mock, it's just a stubbed out implementation of the `TravisChecker` interface. A mock needs to have dynamic behaviour so that we can test different cases like, for example, failing to find a build URL for the repo/ref. 

A pattern to look out for in the future, is when a test case "magically" knows an important detail about the mock implementation. In this case, the `TestTravisFieldIsPopulated` test knows about the magic string `"www.travisci.com"`, it's a sign our test case is really just testing the mock's implementation details.

internal/api/service/service_test.go
```diff
@@ -29,12 +31,28 @@ func newMockK8sClient(name string, time time.Time) *mockK8sClient {
 	}
 }
 
-func TestListReleasesPopulatesFromConfigMap(t *testing.T) {
-	name := "configmap-1"
+type mockGitClient struct {
+	Org    string
```
crdueck: Another red flag, these fields are never used, so why are they here in the first place? The `mockGitClient` should be using the `testify/mock.Mock` type for sure.

internal/api/integrations/k8s/k8s_test.go
```diff
@@ -0,0 +1,87 @@
...
+		{"https://github.com/Wattpad/miranda/", "miranda"},
+		{"https://github.com/Wattpad/", ""},
+		{"https://github.com/Wattpad", ""},
+	}
+	for _, test := range tests {
+		repo, err := findVCSRepo(test.input)
+		if err == nil {
+			assert.Equal(t, test.expected, *repo)
+		} else {
+			assert.Error(t, err)
```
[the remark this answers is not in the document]
rsgc-peterson-e: If I do it this way, test cases that have the function returning an error will fail. Since an error is always returned with an empty string, I will do `assert.Equal(t, test.expected, repo)` where the expected string is also empty.

internal/api/integrations/k8s/k8s.go
```diff
@@ -76,16 +103,46 @@ func (k *K8sClient) ListAll(namespace string) ([]models.Release, error) {
...
-				Github: annotations[annotationFor("code")],
+				Github: *gitRepo,
+				Ref:    image.Tag,
 			},
 			Artifacts: models.Artifacts{
 				Chart: models.HelmArtifact{
 					Path:    r.Spec.Chart.Path,
 					Version: r.Spec.Chart.Revision,
 				},
+				Docker: models.DockerArtifact{
```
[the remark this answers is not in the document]
rsgc-peterson-e: I have added some logic where `image` is assigned, which will ensure the `DockerArtifact` fields are empty string in the case of a `image == nil`. The new code is as follows:

```
		image, err := GetImageForRepo(releaseName, r.Spec.Values)
		if err != nil {
			k.logger.Log("error", err)
		}

		repository := ""
		tag := ""
		if image != nil {
			repository = image.Repository
			tag = image.Tag 
		}
```
...
```
Docker: models.DockerArtifact{
	Image: repository,
	Tag:   tag,
},
```

internal/api/service/service_test.go
rsgc-peterson-e: I have changed the mock git client to use the testify mock library. I have also added a test case for when the travis build URL is not fetched.

internal/api/integrations/k8s/k8s.go
```diff
@@ -67,17 +96,46 @@ func (k *K8sClient) ListAll(namespace string) ([]models.Release, error) {
...
+				Ref:    tag,
 			},
 			Artifacts: models.Artifacts{
 				Chart: models.HelmArtifact{
 					Path:       r.Spec.Chart.Path,
 					Repository: r.Spec.Chart.Repository,
 					Version:    r.Spec.Chart.Revision,
 				},
+				Docker: models.DockerArtifact{
+					Image: repository,
```
crdueck: I think this field should be the absolute path to the image, the `image.URI()`, instead of just the repo, which would be something like `word-counts`.

internal/api/integrations/k8s/k8s.go
```diff
@@ -52,8 +62,27 @@ func (k *K8sClient) ListAll(namespace string) ([]models.Release, error) {
...
+		}
+
+		releaseName := r.GetName()
+
+		image, err := GetImageForRepo(releaseName, r.Spec.Values)
+		if err != nil {
+			k.logger.Log("error", err)
+		}
+
+		repository := ""
```
crdueck: It's personal preference, but explicitly assigning the zero-value smells weird to me in Go. I'd write these like `var repository string`, but there's no significant difference.

cmd/ship-it-api/main.go
```diff
@@ -43,13 +44,14 @@ func main() {
...
 
-	k8s, err := k8s.New(ctx, cfg.HelmReleasesResync())
+	k8s, err := k8s.New(ctx, cfg.HelmReleasesResync(), logger)
 	if err != nil {
 		logger.Log("error", err)
 		os.Exit(1)
 	}
 
-	svc := service.New(k8s)
+	gitClient := github.New(ctx, cfg.GithubOrg, cfg.GithubToken)
```
crdueck: git and github are different things. Even though it's very obvious from context that this client is for github, I'd suggest naming this `githubClient`.

internal/api/integrations/k8s/k8s.go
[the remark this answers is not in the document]
rsgc-peterson-e: Agreed.

internal/image.go
```diff
@@ -27,3 +27,42 @@ func ParseImage(repo string, tag string) (*Image, error) {
 		Tag:        tag,
 	}, nil
 }
+
+func GetImagePath(obj map[string]interface{}, serviceName string) []string {
```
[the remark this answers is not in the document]
rsgc-peterson-e: Since kube-builder is now using the `runtime.RawExtension` type, which @jrhouston has written a helper function for to unmarhshal into a `map[string]interface{}`, I think we can stick to this as it will be compatible with the kube-builder stuff.

internal/api/integrations/k8s/k8s.go
rsgc-peterson-e: Should the `DockerArtifiact` type simply be removed then? If we give the full URI, other information such as the tag and repository name can be parsed out. Perhaps the `Docker` field could just be the URI string. For now I have put the full URI in the `Image` field as requested.

internal/image_test.go
```diff
@@ -29,11 +29,153 @@ func TestParseImage(t *testing.T) {
...
+		assert.Equal(t, test.expected, img)
+	}
+}
 
-	for name, test := range testCases {
-		t.Run(name, func(t *testing.T) {
-			img, _ := ParseImage(test.repo, test.tag)
-			assert.Equal(t, test.expected, img)
-		})
+func TestGetImagePath(t *testing.T) {
```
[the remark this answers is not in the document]
rsgc-peterson-e: Yes I agree, it seems that I made a mistake when moving the code over. This branch now has the same named tests as the `master` branch does.

PR #49, opened 2019-07-12
VELO-1475: syncd

crdueck:
This PR is in a bit of a weird state. It removes code we were using, like ecrconsumer in `cmd/ship-it-syncd/main.go` because that code needs to be updated to match the interfaces defined in `internal/syncd`. It adds code we're not using yet, like some of the config fields in `internal/syncd/config`, because we have blocked PRs depending on this code and I know the unused code will be used in those PRs.

The syncd package provides the image-registry/chart-repo/kube-cluster synchronization logic, but is abstracted to avoid being directly coupled to ECR+SQS/Github+SQS. We still need to think about how we'd choose between different implementations in a configurable way, but for now we can just hardcode our specific implementations. 

Also, I renamed it from `gitsync`, because if you look through the code in `internal/syncd/syncd.go`, there's no reference to git-anything, so it feels weird to be randomly mentioning git in the package name. `syncd` was the best I came up with, since I don't think it's a stretch to imagine it as a background synchronization process.

### DIFF
--- a/internal/image_test.go
+++ b/internal/image_test.go
@@ -7,12 +7,14 @@ import (
 )
 
 func TestParseImage(t *testing.T) {
-	var tests = []struct {
+	type testCase struct {
 		repo     string
 		tag      string
 		expected *Image
-	}{
-		{
+	}
+
+	testCases := map[string]testCase{
+		"valid image": {
 			"foo/bar",
 			"baz",
 			&Image{
@@ -20,14 +22,18 @@ func TestParseImage(t *testing.T) {
 				Repository: "bar",
 				Tag:        "baz",
 			},
-		}, {
+		},
+		"invalid image": {
 			"foo-bar",
 			"baz",
 			nil,
 		},
 	}
-	for _, test := range tests {
-		img, _ := ParseImage(test.repo, test.tag)
-		assert.Equal(t, test.expected, img)
+
+	for name, test := range testCases {
+		t.Run(name, func(t *testing.T) {
+			img, _ := ParseImage(test.repo, test.tag)
+			assert.Equal(t, test.expected, img)
+		})
 	}
 }
--- a/internal/image_test.go
+++ b/internal/image_test.go
@@ -1,0 +1,33 @@
+package internal
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseImage(t *testing.T) {
+	var tests = []struct {
+		repo     string
+		tag      string
+		expected *Image
+	}{
+		{
+			"foo/bar",
+			"baz",
+			&Image{
+				Registry:   "foo",
+				Repository: "bar",
+				Tag:        "baz",
+			},
+		}, {
+			"foo-bar",
+			"baz",
+			nil,
+		},
+	}
+	for _, test := range tests {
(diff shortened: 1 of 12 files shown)


internal/syncd/syncd.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+// ImageReconciler is a callback that reconciles a new image with some external
+// state. For example, updating the values of a helm chart in a remote
+// repository to use the new image's tag.
+type ImageReconciler interface {
+	Reconcile(context.Context, *internal.Image) error
+}
+
+// ImageListener models a stream of `*Image` events. It calls the
```
crdueck: Math trivia: the claim that this interface "models a stream" can be formalized in a stronger type system, by applying a property of [CPS](https://en.wikipedia.org/wiki/Continuation-passing_style) functions over some monad. Practical category theory 🤓

internal/syncd/syncd.go
```diff
@@ -0,0 +1,89 @@
...
+
+// ChartReconciler is an callback that reconciles a new chart with some external
+// state. For example, deploying the chart to a kubernetes cluster.
+type ChartReconciler interface {
+	Reconcile(context.Context, *chart.Chart) error
+}
+
+// ChartListener models a stream of `*chart.Chart` events. It calls the
+// ChartReconciler to reconcile each new chart with some external state.
+type ChartListener interface {
```
jrhouston: Is this the thing that listens for changes to the registry chart?

crdueck: Yes, it listens for changes to the registry chart. But the syncd pkg doesn't need to know or care that it's for a specific chart. Here's how we would implement it with github+sqs: https://github.com/Wattpad/ship-it/pull/50/files#diff-4fd52b3f4241e7dea56c262b4ce0bc67

jrhouston: I feel like this abstraction is going slightly too far. The job of this sync component is always going to be to reconcile between resources in the service registry, images in a docker repository, and the state of the cluster ship-it is deployed to. 

I would prefer that here we explicitly called this `RegistryListener` or `RegistryChartListener` to make the responsibility of this component more clear. I get that we're trying to be abstract to make this pluggable, but I think the tool can prescribe that the registry is going to be a chart.

internal/image_test.go
```diff
@@ -0,0 +1,33 @@
...
+				Repository: "bar",
+				Tag:        "baz",
+			},
+		}, {
+			"foo-bar",
+			"baz",
+			nil,
+		},
+	}
+	for _, test := range tests {
```
jrhouston: Let's use [subtests](https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks) here, see Dave Cheney's article on table driven tests:

https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

internal/syncd/syncd.go
crdueck: `RegistryChartListener` sounds good to me. I updated the docs as well to be less vague about what these interfaces are meant for

jrhouston:
:shipit:

crdueck:
I fixed-up the wrong test cases (but they should also be using sub-tests). One more R pls.

PR #50, opened 2019-07-12
VELO-1452: syncd github integration

crdueck:


internal/syncd/integrations/github/listener.go
```diff
@@ -0,0 +1,77 @@
...
+	stack := sqsmiddleware.ApplyDecoratorsToHandler(
+		l.handler(r),
+		middleware.Timer(l.timer),
+		middleware.Logger(l.logger),
+	)
+	return sqsconsumer.NewConsumer(l.service, stack).Run(ctx)
+}
+
+type pushEvent struct {
+	Ref        string
```
crdueck: I need to add json struct tags for `pushEvent`

internal/syncd/integrations/github/listener.go
```diff
@@ -0,0 +1,77 @@
...
+	svc, err := sqsconsumer.NewSQSService(queue, sqs)
+	if err != nil {
+		return nil, err
+	}
+
+	return &ChartListener{
+		downloader: newDownloader(r, org),
+		logger:     log.With(l, "worker", "github"),
+		service:    svc,
+		timer:      h.With("worker", "github", "queue", queue),
```
crdueck: The `queue` tag is kinda pointless, the `github` worker only uses the single queue so it's redundant. Does it make sense to keep it or should I remove?

rsgc-peterson-e: As we discussed, it is easy to add if needed and can be removed for now.

internal/syncd/integrations/github/listener_test.go
```diff
@@ -0,0 +1,97 @@
...
+func TestHandlerChartDownloadFails(t *testing.T) {
+	errNotFound := errors.New("not found")
+	testEvent := pushEvent{
+		Ref:        "ref",
+		Path:       "path",
+		Repository: "repository",
+	}
+
+	eventBytes, err := json.Marshal(testEvent)
+	if !assert.NoError(t, err) {
```
rsgc-peterson-e: Is this `if` statement necessary? Would the test case fail in the case of an error by just calling `assert.NoError` without explicitly telling it do so through an `if` statement? This pattern is repeated elsewhere so perhaps I am missing something.

internal/syncd/integrations/github/listener_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	return ret0, args.Error(1)
+}
+
+type mockReconciler struct {
+	mock.Mock
+}
+
+func (m *mockReconciler) Reconcile(ctx context.Context, chart *chart.Chart) error {
+	args := m.Called(ctx, chart)
```
rsgc-peterson-e: `m.Called(ctx, chart)` could be inlined into the return statement like so ` return m.Called(ctx, chart).Error(0)`.

internal/syncd/integrations/github/downloader.go
```diff
@@ -0,0 +1,59 @@
...
+
+	var files []*chartutil.BufferedFile
+
+	if file != nil {
+		content, err := file.GetContent()
+		if err != nil {
+			return nil, errors.Wrap(err, "unable to get file contents")
+		}
+
+		files = append(files, &chartutil.BufferedFile{
```
jrhouston: Just a matter of taste but I'd favour returning early here and removing the `else` block below.

internal/syncd/integrations/github/listener_test.go
jrhouston: Wondering this too.

crdueck: `assert.NoError` causes the test case to fail, but doesn't stop the test case execution. If I don't check the return value and call `t.FailNow()`, the test case will continue on and panic or produce a meaningless result because `eventBytes === nil`.

internal/syncd/integrations/github/downloader.go
crdueck: I like it

crdueck:
I forgot that I had my branch for https://github.com/Wattpad/ship-it/pull/53 tracking this branch, and merged that PR into this one. Starting to get a bit unwieldly...

PR #56, opened 2019-07-17
Fix Chart Download Link in UI

rsgc-peterson-e:
VELO-1487
Also adds Repository to release struct which is serialized to json.


web/src/components/GridCell.js
```diff
@@ -83,7 +84,7 @@ class SingleGridCell extends React.Component {
                 {deployDate.toDateString()}
               </div>
               <div className='row-align'>
-                <SelectionDialog source={this.props.SingleGridCellData.code.github} chart={this.props.SingleGridCellData.artifacts.chart.path}/>
+                <SelectionDialog source={this.props.SingleGridCellData.code.github} chart={urljoin('http://' + this.props.SingleGridCellData.artifacts.chart.repository, this.props.SingleGridCellData.artifacts.chart.path)}/>
```
crdueck: We should require that HelmReleases provide a full chart URL including the HTTP scheme, instead of having the frontend add it in.

crdueck: I just updated word-counts: https://github.com/Wattpad/miranda/commit/c6d401141eef10686424c838458f2d8fdda2d080#diff-de84f44dbb021eeefd3a48d8ca892190

rsgc-peterson-e: Agreed, I will remove the leading protocol string.

PR #58, opened 2019-07-17
Convenience methods for getting HelmRelease annotations

crdueck:


internal/api/integrations/k8s/annotations.go
```diff
@@ -0,0 +1,46 @@
...
+type helmReleaseAnnotations map[string]string
+
+func (a helmReleaseAnnotations) get(k string) string {
+	return a[annotationFor(k)]
+}
+
+func (a helmReleaseAnnotations) AutoDeploy() bool {
+	autoDeploy, err := strconv.ParseBool(a.get("autodeploy"))
+	if err != nil {
+		return false
```
crdueck: If we can't parse the annotation, it's not being auto deployed.

PR #60, opened 2019-07-22
Add S3 Downloader Package

rsgc-peterson-e:
VELO-1491


internal/downloader/s3.go
```diff
@@ -0,0 +1,49 @@
...
+
+type Downloader struct {
+	Bucket string
+	d      downloaderAPI
+}
+
+func New(bucketName string, dl downloaderAPI) (*Downloader, error) {
+	if dl == nil {
+		return nil, fmt.Errorf("received nil downloader")
+	}
```
rsgc-peterson-e: Not sure if this check is required, I have included it because the `s3manager.NewDownloader()` function returns a pointer to a downloader without any error. As a result, I added the `if` statement to handle a potential `nil` pointer case.

internal/downloader/s3_test.go
```diff
@@ -0,0 +1,81 @@
...
+	args := m.Called(ctx, w, input)
+	return int64(args.Int(0)), args.Error(1)
+}
+
+func (m *mockS3) DownloadWithIterator(ctx aws.Context, iter s3manager.BatchDownloadIterator, opts ...func(*s3manager.Downloader)) error {
+	args := m.Called(ctx, iter)
+	return args.Error(0)
+}
+
+func TestNewDownloader(t *testing.T) {
```
rsgc-peterson-e: These tests do not do very much besides assert that the `Download()` function behaves as expected. I am open to suggestions on what else could be tested.

internal/downloader/s3.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+)
+
+type downloaderAPI interface {
+	Download(w io.WriterAt, input *s3.GetObjectInput, options ...func(*s3manager.Downloader)) (n int64, err error)
+	DownloadWithContext(ctx aws.Context, w io.WriterAt, input *s3.GetObjectInput, options ...func(*s3manager.Downloader)) (n int64, err error)
```
crdueck: I suggest using `DownloadWithContext` and removing the other interface methods. We only need to depend on methods we actually call

internal/downloader/s3.go
crdueck: As it stands today, `NewDownloader` never returns `nil`: https://github.com/aws/aws-sdk-go/blob/master/service/s3/s3manager/download.go#L79

but, the type signature allows it to possibly return nil in the future. I would personally assume that it will never be nil, and simplify this constructor as a result. But I think either way is reasonable enough.

internal/downloader/s3_test.go
```diff
@@ -0,0 +1,81 @@
...
+		assert.Nil(t, dl)
+	})
+}
+
+func TestDownloadSuccess(t *testing.T) {
+	mockD := &mockS3{}
+	dl, err := New("foo", mockD)
+	assert.NoError(t, err)
+
+	mockD.On("Download", mock.AnythingOfType("*aws.WriteAtBuffer"), &s3.GetObjectInput{
```
crdueck: We can use [Call.Run](https://godoc.org/github.com/stretchr/testify/mock#Call.Run) to perform side-effects, like writing bytes to the `aws.WriteAtBuffer`

internal/downloader/s3.go
```diff
@@ -0,0 +1,49 @@
...
+		return nil, fmt.Errorf("received nil downloader")
+	}
+
+	return &Downloader{
+		Bucket: bucketName,
+		d:      dl,
+	}, nil
+}
+
+func (dl *Downloader) Download(key string) ([]byte, error) {
```
crdueck: We could consider extending this to provide a more complete abstraction. Callers will always want to marshal these bytes into a `*chart.Chart` so they can actually do something with it. We could encapsulate this behaviour in a `S3ChartDownloader` (as an instance of a hypothetical `ChartDownloader` interface defined elsewhere) and provide a more readily usable abstraction.

internal/downloader/s3.go
rsgc-peterson-e: I see, I will go ahead and remove the check for now then.

internal/downloader/s3.go
rsgc-peterson-e: I have added and tested a function which returns the `*chart.Chart` type. As you stated, a general `ChartDownloader` interface can be defined elsewhere if someone would like to use a different chart repository from S3.

internal/downloader/s3.go
```diff
@@ -0,0 +1,70 @@
...
+	})
+
+	if err != nil {
+		return nil, errors.Wrap(err, "chart download failed")
+	}
+
+	return buff.Bytes(), nil
+}
+
+func makeS3Key(chartName string) string {
```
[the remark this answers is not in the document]
rsgc-peterson-e: It was meant to be used when calling the `Download` method from the AWS S3 SDK, just to ensure the key is prefixed with a `/`.

internal/downloader/s3.go
```diff
@@ -0,0 +1,70 @@
...
+
+func New(bucketName string, dl downloaderAPI) Downloader {
+
+	return Downloader{
+		Bucket: bucketName,
+		d:      dl,
+	}
+}
+
+func (dl Downloader) Download(ctx context.Context, key string) ([]byte, error) {
```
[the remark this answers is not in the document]
rsgc-peterson-e: As we discussed, it turns out that the `aws.WriteAtBuffer` type does not have a `Read()` method. Since this is the case, it makes more sense to leave the type signature as is.

internal/downloader/s3_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+func (m *mockS3) DownloadWithContext(ctx aws.Context, w io.WriterAt, input *s3.GetObjectInput, options ...func(*s3manager.Downloader)) (n int64, err error) {
+	args := m.Called(ctx, w, input)
+	return int64(args.Int(0)), args.Error(1)
+}
+
+func TestNewDownloader(t *testing.T) {
+	var m *mockS3
+	dl := New("foo", m)
+	assert.NotNil(t, dl)
```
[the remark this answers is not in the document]
rsgc-peterson-e: Removed.

PR #63, opened 2019-07-26
Add HelmValues method to HelmRelease type

crdueck:
Adding this method reduces code duplication and means that in the future
when we change the representation of the `Spec.Values` field, we'll only
need to update code in one place.

### DIFF
--- a/operator/api/v1beta1/helmrelease_types.go
+++ b/operator/api/v1beta1/helmrelease_types.go
@@ -16,6 +16,8 @@ limitations under the License.
 package v1beta1
 
 import (
+	"encoding/json"
+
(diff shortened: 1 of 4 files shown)


operator/api/v1beta1/helmrelease_types.go
```diff
@@ -58,6 +60,17 @@ type HelmRelease struct {
 	Status HelmReleaseStatus `json:"status,omitempty"`
 }
 
+func (hr *HelmRelease) HelmValues() map[string]interface{} {
```
jrhouston: We should have a test case for this even tho it's temporary.

jrhouston:
:shipit:

PR #64, opened 2019-07-29
List all Images linked to a HelmRelease in Docker UI window

rsgc-peterson-e:
VELO-1511
All images linked to a given release are now listed in the docker dialog
window on the UI

### DIFF
--- a/web/src/components/ImageInfoDialog.js
+++ b/web/src/components/ImageInfoDialog.js
@@ -24,11 +24,11 @@ class ImageInfoDialog extends React.Component {
     render() {
         this.images = this.props.docker.map(d => 
             <div>
-                <ListItem>
+                <ListItem className="ImageListItem">
                     <ListItemText>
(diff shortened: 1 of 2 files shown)


web/src/components/ImageInfoDialog.js
```diff
@@ -1,35 +1,46 @@
...
+    getURI(docker) {
+        return docker.image + ':' + this.getTag()
     }
 
     render() {
+        this.images = this.props.docker.map(d => 
+            <div>
+                <ListItem>
+                    <ListItemText>
+                        <Typography><b>Registry:</b> {this.getRegistry(d)}</Typography>
```
jrhouston: I tend to use <label> and style it rather than a <b> tag for this kind of thing, but this is ok for now.

rsgc-peterson-e: I will push a fix to use a styled `<label>` tag to get the same effect as `<b>`.

rsgc-peterson-e:
Screen Shot 2019-07-29 at 1 44 54 PM

PR #65, opened 2019-07-29
Populate docker artifacts in /releases response

crdueck:
Search the HelmRelease's values to find all `image` sections, then
populate the response object with those values.

This commit introduces the `internal/unstructured` package for working
with unstructured `map[string]interface{}` types, which has come up a
number of times already. I plan to move more related code into this pkg
in later commits.

VELO-1510


internal/unstructured/unstructured.go
```diff
@@ -0,0 +1,19 @@
+// Package unstructured provides utility functions for working with unstructured
+// objects. For example, YAML or JSON data that cannot be strongly typed.
+package unstructured
+
+// FindAll recursively finds every instance of the 'key' in the object, and
+// calls the provided callback function for each value at the matching keys. It
+// will not recurse through the values of a matching key.
+func FindAll(obj map[string]interface{}, key string, cb func(interface{})) {
```
jrhouston: Neat.

internal/unstructured/unstructured_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+type mockCallback struct {
+	mock.Mock
+}
+
+func (m *mockCallback) VoidMethod(x interface{}) {
+	m.Called(x)
+}
+
+func TestFindAll(t *testing.T) {
```
jrhouston: I think this function is simple and pure enough that we can avoid mocks and just have a test case that shows exactly how to use the function:

```go
package unstructured

import (
	"testing"

	"github.com/stretchr/testify/assert"
)

func TestFindAll(t *testing.T) {
	obj := map[string]interface{}{
		"bar": "happy",
		"foo": map[string]interface{}{
			"bar": "path",
		},
		"qux": 2,
	}

	foundValues := []string{}

	FindAll(obj, "bar", func(x interface{}) {
		foundValues = append(foundValues, x.(string))
	})

	assert.ElementsMatch(t, foundValues, []string{"happy", "path"})
}
```

internal/api/integrations/k8s/k8s.go
```diff
@@ -77,9 +76,27 @@ func (k *K8sClient) ListAll(ctx context.Context, namespace string) ([]models.Rel
...
 
 	return releases, nil
 }
+
+func dockerArtifacts(hr shipitv1beta1.HelmRelease) []models.DockerArtifact {
+	var artifacts []models.DockerArtifact
+
+	// find all "image" sections in the release's values, transforming each
+	// one into a docker artifact
+	unstructured.FindAll(hr.HelmValues(), "image", func(x interface{}) {
```
jrhouston: Just being the devil's advocate, but what if there's an `image` value somewhere that isn't a docker image? Should we have a guard here to check if `repository` and `tag` are present?

internal/api/integrations/k8s/k8s_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"testing"
+
+	shipitv1beta1 "ship-it-operator/api/v1beta1"
+	"ship-it/internal/api/models"
+
+	"github.com/stretchr/testify/assert"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func TestDockerArtifacts(t *testing.T) {
```
jrhouston: This is a test for a private function in package. What we should be doing instead is testing the public function exposed by this package `ListAll` and asserting that the docker artifacts get returned.

jrhouston: Testing `ListAll` is going to be slightly hard because we'll need to mock the List method on the k8s client.

Could use this: https://godoc.org/sigs.k8s.io/controller-runtime/pkg/client/fake

crdueck: The `K8sClient` is difficult to unit test because `New()` creates a controller-runtime client from a `rest.InClusterConfig`, depending on a real cluster environment.

I could try to use the [fake client](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/client/fake), but then I'm assigning to the private `K8sClient.client` field, which is related to the same issue you're calling out here.

So before I go forward with the fake client, are we okay with that approach?

crdueck:
uhhh...

https://travis-ci.com/Wattpad/ship-it/jobs/220653191#L525
> internal/api/integrations/k8s/k8s.go:92:25: hr.HelmValues undefined (type v1beta1.HelmRelease has no field or method HelmValues)

but that method definitely exists, and everything builds locally and on the PR test/lint stages
 
https://github.com/Wattpad/ship-it/blob/VELO-1510-docker-artifacts/operator/api/v1beta1/helmrelease_types.go#L63

PR #68, opened 2019-07-30
VELO-1513: HelmRelease image repository informer

crdueck:


internal/syncd/integrations/k8s/informer.go
```diff
@@ -0,0 +1,131 @@
...
+const imageRepositoriesIndex = "ImageRepositoriesIndex"
+
+var errCacheSyncFailed = errors.New("repository informer: image repository cache sync failed")
+
+func errUnexpectedObjectType(obj interface{}) error {
+	return fmt.Errorf("repository informer: unexpected object type %T", obj)
+}
+
+func eventHandler(indexer toolscache.Indexer) *toolscache.ResourceEventHandlerFuncs {
+	// TODO: should we log indexer errors?
```
crdueck: thoughts on this? the `indexer.{Add,Delete}` methods return an error, but it's not really actionable. Do we care enough to log it?

rsgc-peterson-e: It may be good to know that it was these methods that failed when debugging. However, considering the indexer will be called upon often, it could clutter the logs.

internal/syncd/integrations/k8s/informer_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	values := map[string]interface{}{
+		"image": map[string]interface{}{
+			"repository": wordCountsRepository,
+			"tag":        "foo",
+		},
+	}
+
+	valuesRaw, err := json.Marshal(values)
+	if !assert.NoError(t, err) {
```
rsgc-peterson-e: Why not just `assert.NoError(t, err)`? I recall commenting this on another PR, but I believe it was merged before I saw the reason.

crdueck: https://github.com/Wattpad/ship-it/pull/50#discussion_r304529052

rsgc-peterson-e: Makes sense! My bad.

PR #70, opened 2019-08-01
VELO-1484 - /releases/{name}/resources endpoint

crdueck:
I added a `GET /releases/{name}` endpoint as well, since it felt like a surprising omission once we have `GET /releases` & `GET /releases/{name}/resources`


internal/api/service/service.go
```diff
@@ -6,40 +6,47 @@ import (
...
+		return nil, err
 	}
 
-	return releases, err
-}
-
-func (s *Service) getReleaseStatus(_ *models.Release) release.Status_Code {
-	// The default state of a registered service is 'PENDING_INSTALL'.
-	// There's only one possible release status for now.
-	return release.Status_PENDING_INSTALL
```
crdueck: So we used to default any registered service to `PENDING_INSTALL`. The reasoning was that any newly registered service (ie. an associated `HelmRelease` exists in the cluster), would be installed by the operator as soon as it was noticed. However, this meant we couldn't tell if a release hadn't been noticed yet by the operator.

Now a service will be in a `UNKNOWN` state by default, until the operator becomes aware of it and begins to deploy it, at which point it'll update the `HelmReleaseStatus.Code` field to `PENDING_INSTALL`. So we can actually observe that initial state change.

PR #71, opened 2019-08-01
Improve test coverage

crdueck:
I felt inspired to bump up the test coverage after the all-hands. However, I'm not sure how meaningful these tests actually are.

The router tests in particular need to peek inside the `http.Handler` interface so it can reify the routing table, and I'm basically just re-creating the code in `router.go`. The test case ensures we don't ... accidentally delete a handler for an endpoint? Doesn't really feel meaningful, but it does get us high coverage 🙃


jrhouston:
Woo vanity metrics

crdueck:
@tonygoold I wanted to get your thoughts on this. I found writing tests for the routing setup to be pretty awkward. 

The test I ended up writing is straightforward, but I don't think it's a great test. As in, I can't think of a realistic code change we might make that would cause this test to fail. So the tough question for me was how to write a meaningful test for the [router](https://codecov.io/gh/Wattpad/ship-it/src/master/internal/api/router.go#L23)

crdueck:
I had a thought, i'll push an update to this branch soon.

internal/api/router_test.go
```diff
@@ -1,33 +1,34 @@
...
 
-	handler := New(nil, nil)
+func (m *nopService) ListReleases(ctx context.Context) ([]models.Release, error) {
+	return nil, nil
+}
 
-	if routes, ok := handler.(chi.Routes); ok {
-		assertRoutes(t, expectedRoutes, routes)
+func TestRoutes(t *testing.T) {
+	requests := []*http.Request{
```
crdueck: This treats the router as a black box `http.Handler`, but there's some things I dislike about this test compared to what I had previously:

* I need to provide concrete implementations of the router's `Service` and `Histogram` dependencies. These are only relevant to the controller business logic which is thoroughly tested elsewhere. For the router tests I only care about the structure of the routing table, not the handlers.

* it tests less than the previous approach. The `/` route is served by a filesystem handler, so `/index.html` matches but also `/favicon.ico` and dynamic assets like `/static/main.5ee0bde0.chunk.css`. I can't test that requests for all these dynamic paths are handled correctly.

* same issues as originally, what realistic code change would break this test?

crdueck:
Don't want this branch to stick around anymore, so I'm going to merge in the timer middleware test and punt on router tests until I come up with a better strategy.

PR #73, opened 2019-08-07
Autogenerated CRD lands in ship-it chart

jrhouston:
VELO-1490


jrhouston:
Deployment will get added in [VELO-1515](https://wattpad.atlassian.net/browse/VELO-1515)

PR #74, opened 2019-08-07
VELO-1492: Add high level reconcile logic

crdueck:
Create the HelmReleaseReconciler with a helm client so it can reconcile
HelmRelease events by modifying releases. Two method stubs are added to
handle reconciling deleted and updated HelmReleases.

VELO-1492

### DIFF
--- a/operator/controllers/helmrelease_controller.go
+++ b/operator/controllers/helmrelease_controller.go
@@ -23,6 +23,8 @@ import (
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/helm/pkg/helm"
+	"k8s.io/helm/pkg/proto/hapi/chart"
+	rls "k8s.io/helm/pkg/proto/hapi/services"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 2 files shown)


operator/controllers/helmrelease_controller.go
```diff
@@ -44,19 +59,33 @@ func (r *HelmReleaseReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
 }
 
 func (r *HelmReleaseReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&shipitv1beta1.HelmRelease{}).
 		Complete(r)
 }
+
+func (r *HelmReleaseReconciler) onDelete(ctx context.Context, name types.NamespacedName) error {
+	// Update HelmRelease Status to 'PENDING_DELETE' -- but the HelmRelease doesn't exist anymore :(
```
crdueck: Not sure what we'd do here to reflect the `PENDING_DELETE` state, the HelmRelease CR was already deleted by this point

operator/controllers/helmrelease_controller.go
```diff
@@ -44,19 +68,34 @@ func (r *HelmReleaseReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
+	// Delete the release with helm
+	return errNotImplemented
+}
+
+func (r *HelmReleaseReconciler) onUpdate(ctx context.Context, rls shipitv1beta1.HelmRelease) error {
+	// Update HelmRelease Status to 'PENDING_INSTALL' or 'PENDING_UPGRADE'
```
crdueck: It seems like it's a best practice to do sequences of state changes and reconciliation asynchronously over multiple reconciliation events, see `5. Use asynchronous sync loops` [0] and [1] (grep "A very common pattern is to observe events on the resources that we create")

There's some mechanism for managing ownership of cluster objects, so that a reconciliation event for a HelmRelease is queued when a resource it owns is updated [2]. I'm not sure how this works with Helm managing the releases resources, I'm going to take a look at other operators using helm.

[0] https://cloud.google.com/blog/products/containers-kubernetes/best-practices-for-building-kubernetes-operators-and-stateful-apps
[1] https://blog.openshift.com/kubernetes-operators-best-practices/ 
[2] https://godoc.org/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#SetControllerReference

operator/controllers/helmrelease_controller.go
rsgc-peterson-e: As we discussed this morning, if the cluster is in the middle of deleting the release, perhaps we don't really have a need to observe this particular state.

PR #76, opened 2019-08-09
VELO-1492: more operator scaffold

crdueck:


operator/api/v1beta1/helmrelease_types.go
```diff
@@ -26,6 +26,20 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+type HelmReleaseStatusReason string
+
+const (
+	ReasonDeleteError     HelmReleaseStatusReason = "DeleteError"
```
crdueck: I think these status reasons are necessary to be able to handle cycles in the FSM graph of the controller's logic.

For example, we could reach a `FAILED` deployment state by either a failed install (can't rollback), or a failed upgrade (should rollback), or a failed rollback (would be handled specially, since a rollback to a known good version shouldn't ever fail).

operator/controllers/helmrelease_controller.go
```diff
@@ -83,19 +97,28 @@ func (r *HelmReleaseReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-func (r *HelmReleaseReconciler) onUpdate(ctx context.Context, rls shipitv1beta1.HelmRelease) error {
-	// Update HelmRelease Status to 'PENDING_INSTALL' or 'PENDING_UPGRADE'
-	// Attempt the install/upgrade with helm
-	// If it succeeded, set Status to 'DEPLOYED'
-	// Else, set Status to 'PENDING_ROLLBACK'
-	// Attempt the rollback with helm
-	// If it succeeded, set Status to 'DEPLOYED'
-	// Else, set Status to 'FAILED'
-	return errNotImplemented
+func (r *HelmReleaseReconciler) onUpdate(ctx context.Context, rls shipitv1beta1.HelmRelease) (ctrl.Result, error) {
```
crdueck: This returns a `ctrl.Result`, whereas `onDelete` does not, to allow simply re-queuing a delayed reconciliation request after an install/update. This would allow us to implement the desired async behaviour without having to maintain ownership references on the resources created by helm, although it's less robust and efficient overall. Not totally sure which way I'd prefer to go yet

operator/api/v1beta1/helmrelease_types.go
```diff
@@ -40,7 +54,11 @@ type HelmReleaseSpec struct {
 type HelmReleaseStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
-	Code release.Status_Code `json:"code"`
+
+	Code               release.Status_Code     `json:"code"`
+	LastTransitionTime metav1.Time             `json:"lastTransitionTime,omitempty"`
```
jrhouston: Should we model this like other kubernetes resources and put this under a `conditions` field, like how `Deployment` does for example:

```
status:
  conditions:
  - lastTransitionTime: 2019-07-24T20:02:35Z
    lastUpdateTime: 2019-08-09T14:14:03Z
    message: ReplicaSet "track-service-bfd88bdd8" has successfully progressed.
    reason: NewReplicaSetAvailable
    status: "True"
    type: Progressing
  - lastTransitionTime: 2019-08-09T17:06:37Z
    lastUpdateTime: 2019-08-09T17:06:37Z
    message: Deployment has minimum availability.
    reason: MinimumReplicasAvailable
    status: "True"
    type: Available

File: operator/api/v1beta1/helmrelease_types.go [hunk 4]
```diff
@@ -73,6 +91,20 @@ func (hr HelmRelease) HelmValues() map[string]interface{} {
 	return obj
 }
 
+func (hr *HelmRelease) WithStatus(s HelmReleaseStatus) *HelmRelease {
+	var out HelmRelease
+	hr.DeepCopyInto(&out)
+
+	if out.Status.Code == s.Code && out.Status.Reason == s.Reason {
```
jrhouston: Unit test for this conditional?

operator/api/v1beta1/helmrelease_types.go
```diff
@@ -26,6 +26,20 @@ import (
...
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+type HelmReleaseStatusReason string
+
+const (
+	ReasonDeleteError     HelmReleaseStatusReason = "DeleteError"
+	ReasonDeleteSuccess   HelmReleaseStatusReason = "DeleteSuccess"
+	ReasonInstallError    HelmReleaseStatusReason = "InstallError"
+	ReasonInstallSuccess  HelmReleaseStatusReason = "InstallSuccess"
+	ReasonReconcileError  HelmReleaseStatusReason = "ReconcileError"
```
jrhouston: What's this one?

operator/api/v1beta1/helmrelease_types.go
```diff
@@ -26,6 +26,20 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+type HelmReleaseStatusReason string
```
jrhouston: Can we draw a state diagram for this and add it to our docs?

operator/api/v1beta1/helmrelease_types.go
crdueck: Internal controller/reconciler error. It might be useful for handling "impossible" errors, like covering the branch where a `obj.(*HelmRelease)` assertion fails, which should be impossible since this controller only listens for the HelmRelease API.

Maybe it's unnecessary, I just included all the possible reasons for a state change I could think of. I'd remove it if you think YAGNI, easy to add back if we want it again.

operator/api/v1beta1/helmrelease_types.go
crdueck: I like making the HelmRelease's status more similar to other k8s resources, especially since what I had already looks really similar to these conditions. But I'm wondering if multiple conditions makes sense for HelmReleases.

Right now our conditions are basically the release's status code, and those are all mutually exclusive. A release is either FAILED/DEPLOYED/DELETED etc, but would never be in multiple of these states at once. 

Unless we wanted to add the `status: "True"/"False"` fields like your snippet has, but they seem redundant to me. If a condition is true, it will be present in the status. If it's false, it will be removed from the status conditions. Is there a use case they support that I'm not thinking of?

crdueck:
operator-release-states

docs/operator-release-states.dot
```diff
@@ -0,0 +1,33 @@
+digraph fsm {
```
jrhouston: Nice.

operator/api/v1beta1/helmrelease_types.go
jrhouston: I think YAGNI for now, I'd prefer we were pretty explicit here.

PR #80, opened 2019-08-12
UI Polish

rsgc-peterson-e:
VELO-1504


rsgc-peterson-e:
Screen Shot 2019-08-12 at 12 00 39 PM
Screen Shot 2019-08-12 at 11 55 33 AM

web/src/App.js
```diff
@@ -6,7 +6,8 @@ import axios from 'axios';
 import { CircularProgress } from '@material-ui/core';
 import urljoin from 'url-join'
 
-const API_ADDRESS = urljoin(window.location.protocol + '//' + window.location.host, 'api')
+// const API_ADDRESS = urljoin(window.location.protocol + '//' + window.location.host, 'api')
+const API_ADDRESS = 'http://localhost:8080/api'
```
crdueck: This shouldn't be committed right? Also, why do we need to change the API address now that we have skaffold for local testing?

web/src/components/GridCell.js
```diff
@@ -32,7 +33,12 @@ class SingleGridCell extends React.Component {
   }
 
   cellClick = (event) => {
-    this.props.handleCellClick(event)
+    // this.props.handleCellClick(event)
```
crdueck: Is this commented code still needed? If not, we should remove it

web/src/App.js
rsgc-peterson-e: My bad, I will remove the test URL from the commit. I still use a local test server to see my changes in real time as the `npm build` and `skaffold` can take a long time.

web/src/components/GridCell.js
rsgc-peterson-e: Will remove.

PR #84, opened 2019-08-14
Fix Infinite Loop in ECR Queue Consumer

rsgc-peterson-e:
VELO-1529

The consumer returns an error when the image tag is not a valid commit
hash (ie. "latest"). The problem with this behaviour is that the SQS
consumer library believes the message has not been successfully
processed because the handler returned an error. As a result, the
invalid image is put back on the queue being processed over and over in an
endless loop. This PR has the handler return `nil` on these types of
images so they are removed from the queue. The invalid image tag error is still logged using the logger attached to the `ImageListener` type.

### DIFF
--- a/internal/syncd/integrations/ecr/listener.go
+++ b/internal/syncd/integrations/ecr/listener.go
@@ -3,6 +3,7 @@ package ecr
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"regexp"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/syncd/integrations/ecr/listener.go
```diff
@@ -72,7 +71,8 @@ func (l *ImageListener) handler(r syncd.ImageReconciler) sqsconsumer.MessageHand
 		}
 
 		if !validImageTagRegex.MatchString(ecrEvent.Tag) {
-			return fmt.Errorf("malformed image tag")
+			l.logger.Log("error", "invalid image tag, the image will be ignored")
```
crdueck: This log message doesn't contain enough info to ever be actionable. Let's mention the name of the repo that contains the invalid image tag, as well as the invalid tag itself. A message like `ignoring event for invalid image tag "latest" found in repo "repo"` would be much more actionable.

PR #85, opened 2019-08-14
VELO-1527: operator install 

crdueck:


operator/chartdownloader/factory.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+	"k8s.io/helm/pkg/proto/hapi/chart"
+)
+
+type ChartDownloader interface {
+	Download(context.Context, string) (*chart.Chart, error)
+}
+
+// TODO: github, GCS support?
+type Providers interface {
```
crdueck: I'm open to ideas for how to do this differently/better. Dependency injection for a factory-thing is awkward, since you have to provide all the dependencies for every possible output option. This interface/struct-of-funcs pattern allows only defining the ones you expect to use, and missing providers are handled gracefully, but I'm sure it could be improved.

crdueck:
This test error is coming from `onDelete`: https://travis-ci.com/Wattpad/ship-it/jobs/225303881#L774

For whatever reason, the `DeletionTimestamp` is non-nil and so we call `onDelete` unexpectedly. I could comment out the [branch](https://github.com/Wattpad/ship-it/pull/85/files#diff-a76fcebd9b2cb18f81fa0b2611d79357R124) to make the tests pass, but I'd really like to know what's going on.

operator/controllers/helmrelease_controller.go
```diff
@@ -115,7 +153,54 @@ func (r *HelmReleaseReconciler) onDelete(ctx context.Context, rls shipitv1beta1.
...
+		return ctrl.Result{}, errors.Wrapf(err, "failed to install release %s using chart %s", releaseName, chartURI)
+	}
+
+	rls.Status.SetCondition(shipitv1beta1.HelmReleaseCondition{
+		Type:    release.Status_PENDING_INSTALL.String(),
+		Message: fmt.Sprint("installing chart", chartURI),
+	})
+
+	if err := r.Update(ctx, &rls); err != nil {
+		r.Log.Info("failed to update HelmRelease status", "release", releaseName, "status", release.Status_PENDING_INSTALL)
```
crdueck: There's a bit of chicken-egg here. Do we set the HelmRelease status to PENDING_INSTALL before calling helm (what if the helm call fails?) or do we call helm before setting PENDING_INSTALL (what if the HelmRelease update fails?)?

jrhouston:
As discussed in person:

- [ ] Need to add `omitempty` to the Conditions field of the HelmRelease
- [ ] Test is broken because Delete is being called int the JustBeforeEach function but Delete isn't implemented

operator/main.go
```diff
@@ -42,9 +47,20 @@ func init() {
...
+		gracePeriod          time.Duration
+		metricsAddr          string
+		namespace            string
+		enableLeaderElection bool
+	)
+
+	flag.StringVar(&awsRegion, "aws-region", "", "The AWS region where the operator's chart repository is hosted")
+	flag.StringVar(&chartRepository, "chart-repository", "", "The URI of the chart repository used by the operator")
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&namespace, "namespace", "default", "The cluster namespace where the operator will deploy releases")
```
jrhouston: Why do we need to be explicit about this? Can't we just use `default` unless it's specified in the chart?

operator/controllers/helmrelease_controller.go
```diff
@@ -115,7 +154,54 @@ func (r *HelmReleaseReconciler) onDelete(ctx context.Context, rls shipitv1beta1.
...
+	rls.Status.SetCondition(shipitv1beta1.HelmReleaseCondition{
+		Type:    release.Status_PENDING_INSTALL.String(),
+		Message: fmt.Sprintf("installing chart %s", chartURI),
+	})
+
+	if err := r.Update(ctx, &rls); err != nil {
+		r.Log.Info("failed to update HelmRelease status", "release", releaseName, "status", release.Status_PENDING_INSTALL)
+	}
+
+	return ctrl.Result{RequeueAfter: r.GracePeriod}, nil
```
jrhouston: I'm not 100% clear on how this works. What will happen if we call this, then the controller pod dies before the grace period?

operator/controllers/helmrelease_controller.go
jrhouston: We should be calling helm then setting `PENDING_INSTALL`. I don't think we need to be so discrete as to represent the state of the controller having become aware that it needs to run install but hasn't done so yet.

operator/controllers/helmrelease_controller_test.go
```diff
@@ -0,0 +1,157 @@
...
+			},
+			Spec: shipitv1beta1.HelmReleaseSpec{
+				ReleaseName: releaseName,
+				Chart: shipitv1beta1.ChartSpec{
+					Repository: "github.com/example/foo",
+					Path:       "bar/baz",
+				},
+				Values: runtime.RawExtension{Raw: []byte("{}")},
+			},
+			Status: shipitv1beta1.HelmReleaseStatus{
```
jrhouston: Can just omit this whole thing.

operator/controllers/helmrelease_controller_test.go
```diff
@@ -0,0 +1,157 @@
...
+		reconciler = NewHelmReleaseReconciler(log, k8sClient, helmClient, downloader)
+
+		testRelease = &shipitv1beta1.HelmRelease{
+			TypeMeta: metav1.TypeMeta{
+				Kind: "HelmRelease",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:              releaseName,
+				Namespace:         releaseNamespace,
+				DeletionTimestamp: nil,
```
jrhouston: Can just omit this.

operator/main.go
```diff
@@ -42,9 +47,20 @@ func init() {
 }
 
 func main() {
-	var metricsAddr string
-	var enableLeaderElection bool
+	var (
+		awsRegion            string
```
jrhouston: We'll need to add tillerAddr here too and pass it down.

operator/main.go
crdueck: AFAIK charts don't specify the namespace they should be deployed in. That's a decision we make at install time. We could add a namespace field to the `HelmReleaseSpec` if we wanted, but otherwise this is the only other place we choose which namespace to deploy in.

operator/controllers/helmrelease_controller_test.go
crdueck: yep, had it set while trying to debug that deletion issue

operator/controllers/helmrelease_controller_test.go
crdueck: I'm not able to get the tests to run if `Conditions == nil`, with or without the `omitempty` struct tag. I get a k8s client error `validation failure list:\nstatus.conditions in body is required"`

operator/controllers/helmrelease_controller.go
crdueck: I'm not sure either tbh. I don't know if the controller persists its request queue, but I expect it's probably in memory. 

I'd like to find the answer to that, but I think the best solution would be to eventually implement the "watch resources created by the chart" TODO in this func which completely avoids the need to `RequeueAfter`

operator/controllers/helmrelease_controller_test.go
```diff
@@ -0,0 +1,158 @@
...
+			Expect(k8sClient.Create(ctx, testRelease)).To(Succeed())
+
+			res, err := reconciler.Reconcile(request)
+			Expect(err).To(BeNil())
+			Expect(res).To(BeZero())
+		})
+	})
+
+	When("the HelmRelease has autodeploy enabled", func() {
+		It("should reconcile the release", func() {
```
crdueck: I had to re-organize the tests so that the `BeforeEach/AfterEach` worked correctly. I thought they would wrap each `When` block, but it's actually the `It` blocks. This reads a bit worse that the previous iteration, I think, but it's the best I came up with.

operator/chartdownloader/factory.go
jrhouston: We can worry about this later.

PR #88, opened 2019-08-19
Refactor chartdownloader & fixups

crdueck:
The chartdownloader was picking a chart repository at application start
up, instead of using the values from the HelmRelease being reconciled.
The S3 downloader now correctly parses the bucket and object from the
chart URL.

### DIFF
--- a/operator/chartdownloader/factory.go
+++ b/operator/chartdownloader/factory.go
@@ -4,51 +4,34 @@ import (
 	"context"
 	"fmt"
 	"net/url"
-	"strings"
 
-	"github.com/aws/aws-sdk-go/aws/client"
-	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+	"github.com/pkg/errors"
 	"k8s.io/helm/pkg/proto/hapi/chart"
 )
 
-type ChartDownloader interface {
+type Interface interface {
 	Download(context.Context, string) (*chart.Chart, error)
 }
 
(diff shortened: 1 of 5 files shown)


operator/chartdownloader/factory.go
```diff
@@ -4,51 +4,34 @@ import (
...
-	"strings"
 
-	"github.com/aws/aws-sdk-go/aws/client"
-	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+	"github.com/pkg/errors"
 	"k8s.io/helm/pkg/proto/hapi/chart"
 )
 
-type ChartDownloader interface {
+type Interface interface {
```
crdueck: I dislike this naming convention, but dislike `chartdownloader.ChartDownloader` even more. I've seen `Interface` being used in helm/k8s libs so 🤷‍♂

operator/chartdownloader/s3_test.go
```diff
@@ -27,52 +31,14 @@ func (m *mockS3) DownloadWithContext(ctx aws.Context, w io.WriterAt, input *s3.G
 	return int64(args.Int(0)), args.Error(1)
 }
 
-func TestDownloadSuccess(t *testing.T) {
```
crdueck: I removed these tests of the private `S3Downloader.download` method. It's called directly by the `S3Downloader.Download) method, which is tested.

PR #90, opened 2019-08-19
HelmRelease delete logic

jrhouston:


jrhouston:
bunch of go mod vendor noise in this PR.

operator/controllers/helmrelease_controller_test.go
```diff
@@ -151,8 +151,14 @@ var _ = Describe("HelmReleaseReconciler", func() {
 			By("reconciling a failed updated release")
 			// TODO
 
-			By("reconciling a release with the deletion timestamp set ")
-			// TODO
+			By("reconciling a release that has been deleted")
```
jrhouston: I don't really like that this test depends on the previous one, but I guess this is testing the full lifecycle.

crdueck: The alternative would be a sequence of tests like

* install success/fail
* install success, update success/fail
* install success, update fail, rollback success/fail

etc, each test case has to set up more and more initial state. I didn't think that would scale well, so I went for a full lifecycle single test case.

operator/config/samples/shipit_v1beta1_helmrelease.yaml
```diff
@@ -4,7 +4,6 @@ kind: HelmRelease
 metadata:
   name: example-microservice
   annotations:
-    helmreleases.shipit.wattpad.com/autodeploy: "true"
```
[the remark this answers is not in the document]
jrhouston: Huh, we do - not sure how that got removed.

operator/controllers/helmrelease_controller.go
```diff
@@ -148,8 +148,59 @@ func (r *HelmReleaseReconciler) setFinalizer(ctx context.Context, rls shipitv1be
...
+		return ctrl.Result{}, err
+	}
+
+	info := rlsStatus.GetInfo()
+
+	if info.Status.Code == release.Status_DELETING {
+		return ctrl.Result{RequeueAfter: r.GracePeriod}, nil
+	}
+
+	if info.Status.Code == release.Status_DELETED {
```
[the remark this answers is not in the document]
jrhouston: Yeah I think when it hits the deleted state we just remove the HelmRelease resource.

operator/controllers/helmrelease_controller.go
```diff
@@ -148,8 +148,59 @@ func (r *HelmReleaseReconciler) setFinalizer(ctx context.Context, rls shipitv1be
...
+		Type:    release.Status_DELETING.String(),
+		Message: "Release is being deleted",
+	})
+
+	if err := r.Update(ctx, &rls); err != nil {
+		r.Log.Error(err, "unable to update status")
+		return ctrl.Result{}, err
+	}
+
+	_, err = r.helm.DeleteRelease(rls.Spec.ReleaseName)
```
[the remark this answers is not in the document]
jrhouston: My bad, will fix

crdueck:
I just noticed that this PR didn't change any imports, but lots of vendored deps are being updated. Seems odd. Did you run `go mod tidy` on this branch? If not, does it change the diff at all?

jrhouston:
Oh it's because I ran go mod vendor inside the operator directory

operator/controllers/helmrelease_controller_test.go
jrhouston: Yeah I'm okay with this for now.

operator/api/v1beta1/helmrelease_types.go
```diff
@@ -31,7 +31,6 @@ type HelmReleaseStatusReason string
 
 const (
 	ReasonDeleteError     HelmReleaseStatusReason = "DeleteError"
```
crdueck: The DeleteError const can be removed as well.

jrhouston: Actually, maybe we want to keep that one and set a failure condition here: https://github.com/Wattpad/ship-it/pull/90/files#diff-a76fcebd9b2cb18f81fa0b2611d79357R190

jrhouston: I'll leave as is for now.

PR #96, opened 2019-08-21
VELO-1515: vendorless docker build

crdueck:
This PR adds some dockerfile optimizations for ship-it-{api,syncd} and removes the vendor directory from git. Version-controlling the vendor dir causes a lot of code churn, and isn't necessary when go module are used properly. Build times are about the same, as we can have travis cache the modules and go caches locally already.


jrhouston:
Good riddance

PR #97, opened 2019-08-21
S3 chart downloader should create the full URL to the chart tgz

jrhouston:
VELO-1550

Changed a couple of things here, see comments.


operator/Makefile
```diff
@@ -10,7 +10,7 @@ all: manager
 
 # Run tests
 test: generate fmt vet manifests
-	go test ./api/... ./controllers/... -coverprofile cover.out
+	go test ./api/... ./controllers/... ./chartdownloader/... -coverprofile cover.out
```
jrhouston: This wasn't even running in CI 😬

operator/api/v1beta1/helmrelease_types.go
```diff
@@ -69,12 +69,12 @@ type HelmReleaseCondition struct {
 // ChartSpec defines the desired Helm chart
 type ChartSpec struct {
 	Repository string `json:"repository"`
-	Path       string `json:"path"`
-	Revision   string `json:"revision"`
+	Name       string `json:"name"`
```
jrhouston: Looking at the helm docs, they refer to the chart URL as being made up of a repository and a name, and then you can supply a version as a flag. Just want to mirror this here.

operator/main.go
```diff
@@ -56,10 +56,10 @@ func main() {
 		enableLeaderElection bool
 	)
 
-	flag.StringVar(&awsRegion, "aws-region", "", "The AWS region where the operator's chart repository is hosted")
+	flag.StringVar(&awsRegion, "aws-region", "us-east-1", "The AWS region where the operator's chart repository is hosted")
```
jrhouston: This is a sane default.

operator/main.go
```diff
@@ -56,10 +56,10 @@ func main() {
...
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&namespace, "namespace", "default", "The cluster namespace where the operator will deploy releases")
-	flag.StringVar(&tillerAddr, "tiller-address", "", "The cluster address of the tiller service")
+	flag.StringVar(&tillerAddr, "tiller-address", "localhost:44134", "The cluster address of the tiller service")
```
jrhouston: This should probably change to the full kube dns name for tiller.

operator/Makefile
crdueck: I think we should just use `go test ./...` , so we don't have to keep adding new pkgs

operator/chartdownloader/factory.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 type ChartDownloader interface {
-	Download(context.Context, string) (*chart.Chart, error)
+	Download(context.Context, string, string) (*chart.Chart, error)
```
crdueck: I tend to omit param names when there's only one param per type. But since there's now 2 `string` args, I think param names should be added to make it clear that the chart url comes first, then the version. 

```suggestion
	Download(ctx context.Context, chart string, version string) (*chart.Chart, error)
```

jrhouston:
Forgot to update API server, will do

operator/Makefile
jrhouston: Sold

operator/api/v1beta1/helmrelease_types.go
```diff
@@ -69,12 +69,12 @@ type HelmReleaseCondition struct {
...
+	Version    string `json:"version"`
 }
 
-func (cs *ChartSpec) URI() string {
-	return path.Join(cs.Repository, cs.Path)
+func (c ChartSpec) URL() string {
+	return fmt.Sprintf("%s/%s", c.Repository, c.Name)
```
crdueck: We could `TrimSuffix(c.Repository, "/")` to handle an edge case like `s3://helm.wattpad.com/`. Since the chart now has a `Name` instead of `Path` i don't think we would have to worry about leading/trailing slashes there.

operator/controllers/helmrelease_controller.go
```diff
@@ -266,23 +266,24 @@ func (r *HelmReleaseReconciler) update(ctx context.Context, rls shipitv1beta1.He
...
 	if err != nil {
-		return ctrl.Result{}, errors.Wrapf(err, "failed to download chart %s", chartURI)
+		return ctrl.Result{}, errors.Wrapf(err, "failed to download chart %s", chartURL)
 	}
 
 	// TODO: use the returned response's `Release.Manifest` to watch and
 	// receive events for the k8s resources owned by this chart
-	if _, err := r.helm.InstallReleaseFromChart(chart, r.Namespace, helm.ReleaseName(releaseName)); err != nil {
-		return ctrl.Result{}, errors.Wrapf(err, "failed to install release %s using chart %s", releaseName, chartURI)
+	if _, err := r.helm.InstallReleaseFromChart(chart, r.Namespace, helm.ReleaseName(releaseName), helm.ValueOverrides(rls.Spec.Values.Raw)); err != nil {
```
jrhouston: This resolves VELO-1554

PR #100, opened 2019-08-23
VELO-1532: reconciler update

crdueck:
There's a few related things in here, involving getting the status updates to not endlessly trigger reconciliation requests.

I verified that the unit tests pass and ran a local integration test where I installed and updated an example release.


operator/controllers/helmrelease_controller.go
```diff
@@ -244,71 +246,127 @@ func (r *HelmReleaseReconciler) update(ctx context.Context, rls shipitv1beta1.He
...
 		Type:    release.Status_PENDING_ROLLBACK.String(),
-		Message: fmt.Sprintf("rolling back %s", rls.Spec.ReleaseName),
+		Message: fmt.Sprintf("rolling back %s", releaseName),
+	})
+
+	r.Log.Info("HelmRelease rolled back", "release", releaseName)
+	return ctrl.Result{}, r.Status().Update(ctx, &rls)
+}
+
+func (r *HelmReleaseReconciler) upgrade(ctx context.Context, rls shipitv1beta1.HelmRelease) (ctrl.Result, error) {
```
crdueck: This is almost a copy-paste of the `install` method, idk if there's a good way to DRY

jrhouston:
As discussed at retro, we will merge this and bugfix the logic in new branches.

PR #103, opened 2019-09-03
VELO-1496: release lifecycle events

crdueck:
I added a `ReleaseManager` type to encapsulate release lifecycle operations and events. The intent is for it to manage every direct change to a `HelmRelease` struct such as updating the status or finalizers, as well as calling CRUD methods on the Helm client. It also ensures that events are consistently sent for each status update without adding repetitive code to the reconciler.

The reconciler decides which lifecycle operation it should perform to reconcile the release, then delegates to the `ReleaseManager` to do the op, and lastly uses its own `controller-runtime.Client` to sync the updated `HelmRelease` with the cluster.


operator/controllers/release_manager.go
```diff
@@ -0,0 +1,192 @@
...
+
+func (m *ReleaseManager) updateCondition(cond shipitv1beta1.HelmReleaseCondition) *shipitv1beta1.HelmRelease {
+	m.release.Status.SetCondition(cond)
+	// "Normal" is the default kubernetes event type
+	m.recorder.Event(&m.release, "Normal", cond.Type, cond.Message)
+
+	return m.release.DeepCopy()
+}
+
+func (m *ReleaseManager) ChartSpec() *shipitv1beta1.ChartSpec {
```
crdueck: I chose to add these 3 getter methods rather than making `ReleaseManager.release` a public field, so that it's not possible to mutate the `ReleaseManager`'s internal release state through one of the returned pointers.

operator/controllers/release_manager.go
```diff
@@ -0,0 +1,195 @@
...
+	helm     HelmClient
+	recorder record.EventRecorder
+	release  shipitv1beta1.HelmRelease
+}
+
+func (m *ReleaseManager) updateCondition(cond shipitv1beta1.HelmReleaseCondition) *shipitv1beta1.HelmRelease {
+	m.release.Status.SetCondition(cond)
+	m.recorder.Event(&m.release, v1.EventTypeNormal, cond.Type, cond.Message)
+
+	return m.release.DeepCopy()
```
crdueck: `DeepCopy` here so that the `ReleaseManager`'s internal release state can't be mutated through a returned `*shipitv1beta1.HelmRelease` pointer

operator/controllers/helmrelease_controller.go
```diff
@@ -36,24 +37,30 @@ import (
...
 type ChartDownloader interface {
 	Download(ctx context.Context, chart string, version string) (*chart.Chart, error)
 }
 
 type HelmClient interface {
+	ReleaseCRUD
+	ReleaseStatuser
+}
+
+type ReleaseCRUD interface {
```
crdueck: I liked separating the `HelmClient` methods used by the reconciler vs the release manager, though there's probably a better name for this interface.

PR #104, opened 2019-09-04
operator should re-use helm release name if it was deleted but not purged

jrhouston:
VELO-1564


jrhouston:
Did a manual integration test locally and this works.

operator/controllers/helmrelease_controller.go
```diff
@@ -306,7 +308,7 @@ func (r *HelmReleaseReconciler) install(ctx context.Context, rls shipitv1beta1.H
 
 	// TODO: use the returned response's `Release.Manifest` to watch and
 	// receive events for the k8s resources owned by this chart
-	if _, err := r.helm.InstallReleaseFromChart(chart, r.Namespace, helm.ReleaseName(releaseName), helm.ValueOverrides(rls.Spec.Values.Raw)); err != nil {
+	if _, err := r.helm.InstallReleaseFromChart(chart, r.Namespace, helm.ReleaseName(releaseName), helm.InstallReuseName(true), helm.ValueOverrides(rls.Spec.Values.Raw)); err != nil {
```
crdueck: ahh true, I guess `ReuseName` does what you'd expect when there's no name to re-use.

PR #105, opened 2019-09-05
add Makefile target to spin up a kind cluster with tiller installed

jrhouston:
Just adding this because I'm going to start using kind over minikube.


testdata/tiller_rbac.yaml
```diff
@@ -0,0 +1,19 @@
+---
```
crdueck: I expect `testdata` to contain static inputs for unit tests, so this file surprised me somewhat.

A pattern I've seen in [other](https://github.com/operator-framework/operator-sdk/tree/master/hack) [popular](https://github.com/etcd-io/etcd/tree/master/hack) [projects](https://github.com/moby/moby/tree/master/hack) is to have a top-level `hack` directory containing dev-only resources. We could have `hack/kind/tiller-rbac.yaml`, for example.

Another alternative [suggested](https://github.com/golang-standards/project-layout/tree/master/test) by the golang-standards project-layout is a top-level `test` dir, then we could have `test/tiller-rbac.yaml` or `test/kind/tiller-rbac.yaml`, etc.

I'd prefer either of these alternatives over putting this RBAC in the `testdata` dir, what do you think?

jrhouston: Yeah I'll make a `hack` directory.

PR #108, opened 2019-09-10
Add HelmRelease rollback tests

crdueck:
The FakeClient we're using to mock helm is missing some crucial
functionality of the default client. For example, its `Rollback` method
is a no-op, and the `Delete` method matches the behaviour of `helm
delete --purge`.

In the short term, the open internals of the FakeClient allow us to fake
whatever we need to, though it does start to defeat the purpose of the
test. Hopefully a mediocre test is better than no test at all.


crdueck:
not sure what's up with codecov

PR #111, opened 2019-09-17
UI: Implement searching by release name, owning squad or status

rashadsookram:
This PR adds a search bar in the top bar of the web UI which allows people to filter the list of releases by name, owning squad or status. It's implemented by a simple string `includes` on each field to check if any field contains the query.

After searching, pressing the `x` that appears on the right of the search bar (or pressing escape) will clear the query, showing all the releases again.

Since this is my first PR to this repo, let me know if I'm missing anything. I checked the tests for the web UI, and it looks like there's only one really basic test. If we're going to start adding tests, let me know and I can add some to this PR.

## Screenshots

Screen Shot 2019-09-17 at 10 09 50 AM
Screen Shot 2019-09-17 at 10 09 55 AM
Screen Shot 2019-09-17 at 10 10 01 AM

VELO-1547

### DIFF
--- a/web/src/App.css
+++ b/web/src/App.css
@@ -65,7 +65,7 @@
 }
 
 .search-releases {
-  width: 250px;
+  width: 300px;
   margin-left: 24px;
 }
 
--- a/web/src/App.css
+++ b/web/src/App.css
@@ -64,6 +64,10 @@
   margin: auto;
 }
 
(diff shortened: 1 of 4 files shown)


jrhouston:
This is great, thanks @rashadsookram 😄

PR #118, opened 2019-10-01
Reduce ship-it-api build time by caching more

crdueck:
`npm install react-scripts` can take dozens of seconds to complete. The way it was written originally, any source code changes cause the layer that installs the scripts to be uncached. Re-organizing allows better re-use of cached layers and avoids re-downloading the same package during local dev.


crdueck:
Apparently `react` isn't installed as a dependency of `react-scripts`... updating PR

crdueck:
I re-ordered things to install the `package.json` dependencies in a container layer before we depend on the rest of the web UI source code.

PR #120, opened 2019-10-02
VELO-1581: install ship-it to kind for local dev

crdueck:
Would appreciate someone trying the `Local Development` steps in the readme and letting me know if there's any bumps along the way


Makefile
```diff
@@ -41,8 +41,16 @@ kind:
...
 	@echo Done! Set your kubectl context:
 	@echo
 	@echo export KUBECONFIG=$(KUBECONFIG)
+
+kind-install-ship-it:
+	kind load docker-image --name $(KIND_CLUSTER_NAME) ship-it-api:$(VERSION)
+	kind load docker-image --name $(KIND_CLUSTER_NAME) ship-it-syncd:$(VERSION)
+	kind load docker-image --name $(KIND_CLUSTER_NAME) ship-it-operator:$(VERSION)
+	helm upgrade --install localstack deploy/localstack
+	helm upgrade --install ship-it deploy/ship-it --set api.image.tag=$(VERSION),api.image.repository=ship-it-api,syncd.image.tag=$(VERSION),syncd.image.repository=ship-it-syncd,operator.image.tag=$(VERSION),operator.image.repository=ship-it-operator,devEnv.DOGSTATSD_HOST="localhost"
```
crdueck: We could have a `value-dev.yaml` or something, I think there's a way to merge override values on the command line. But this also works, any preference one way or the other?

README.md
```diff
@@ -8,41 +8,54 @@ The technical background and architecture of Ship-it can be found in the overvie
...
 ```bash
-$ minikube update-context
+$ export KUBECONFIG=$(kind get kubeconfig-path --name="ship-it-dev")
 $ kubectl config current-context
-minikube
+kubernetes-admin@ship-it-dev
 ```
 
-3. Install tiller in minikube
+3. Build local images
```
crdueck: It would be great to automate this instead of having it as a manual step

hack/ship-it-dev.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: kind.sigs.k8s.io/v1alpha3
+kind: Cluster
+metadata:
+  name: ship-it-dev
+nodes:
+  - role: control-plane
+    extraPortMappings:
+      - hostPort: 31901
```
crdueck: This'll break if the nodePort changes :( not sure if there's a good way to keep them in-sync. Not too big a deal though

jrhouston: Yeah this is OK for now.

Makefile
jrhouston: I'm okay with doing it this way

README.md
crdueck: This step would be a bit nicer if there was a single `make images` (or whatever) target instead of 3 different ones. The env target thing was neat but it's starting to annoy me in practice. Might try to get it down to a single target, even if there's more duplication in the makefile.

PR #121, opened 2019-10-18
Add Makefile target to publish ship-it chart

crdueck:
The target depends on the user already having helm + helm-s3 plugin
installed and having added the 's3://charts.wattpadhq.com' repository to
their repo list.

https://github.com/hypnoglow/helm-s3

---

Adding a convenience target to publish the chart, so we can use it in the cluster-components chart in miranda.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,8 @@
 .PHONY: build docs jsonschema push
 
+CHART_VERSION := $(shell helm inspect deploy/ship-it | awk '/version/{ print $$2; }')
+CHART_REPOSITORY := $(shell helm repo list | awk '/wattpad/{ print $$2; }')
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -22,6 +25,10 @@ push: build
 	docker push $(REGISTRY)/$(TARGET_IMAGE)
 	docker push $(REGISTRY)/$(LATEST_IMAGE)
 
+chart:
+	helm package deploy/ship-it
+	helm s3 push ship-it-$(CHART_VERSION).tgz $(CHART_REPOSITORY)
```
jrhouston: Helm s3 plugin needs to be installed for this to work

crdueck:
The travis error is from the nodejs build:

> Cannot find module: '@material-ui/core/styles/MuiThemeProvider'. Make sure this package is installed.

I'm gonna go ahead and merge this PR

PR #124, opened 2019-11-14
Add ExternalSecret to ship-it chart

crdueck:
We should eventually de-wattpad the chart's default values and configure it all in miranda. I commented something about this on VELO-1589 in jira. Then we could have a `prod-v3/values.yaml` without external secret, and a `prod-v4/values.yaml` using external secret.

### DIFF
--- a/deploy/ship-it/values.yaml
+++ b/deploy/ship-it/values.yaml
@@ -2,7 +2,7 @@ nameOverride: ""
 fullnameOverride: ""
 
 awsRegion: us-east-1
-sslCertPath: /etc/ssl/certs/ca-certificates.crt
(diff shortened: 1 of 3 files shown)


deploy/ship-it/values.yaml
```diff
@@ -2,12 +2,14 @@ nameOverride: ""
 fullnameOverride: ""
 
 awsRegion: us-east-1
-existingSecretName: ship-it
-sslCertPath: /etc/ssl/certs/ca-certificates.crt
+sslCertPath: /etc/ssl/certs/ca-bundle.crt
```
crdueck: This host path has changed in EKS linux AMI compared to what kops was using.

PR #129, opened 2019-11-26
Speed up Travis build

tomtomyan:
Cuts build time from 12 minutes to roughly 7.5 minutes

- Build stage jobs run immediately after testing jobs so it re-uses cached packages
- Codecov to run in Test and Lint stage so we aren't running `go test` twice
- Use recommended and faster method for installing golangci-lint

VELO-1771

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ jobs:
   include:
     - stage: Test and Lint
       name: Go and Codecov
-      install: GO111MODULE=off go get "github.com/golangci/golangci-lint/cmd/golangci-lint"
+      install: curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin v1.21.0
       script:
         - golangci-lint --deadline=5m run -D errcheck -E goimports -E interfacer -E scopelint ./...
         - go test -race -coverprofile=coverage.txt -covermode=atomic ./...
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,15 +46,6 @@ jobs:
       name: JavaScript
       script: cd web && npm install -D && npm run lint && npm run test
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -16,34 +16,20 @@ services:
...
-        - cd operator
-      script:
-        - make docker-build
-
     - stage: Test and Lint
       name: Go
-      install: GO111MODULE=off go get "github.com/golangci/golangci-lint/cmd/golangci-lint"
+      install: curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin v1.21.0
       script:
         - golangci-lint --deadline=5m run -D errcheck -E goimports -E interfacer -E scopelint ./...
```
crdueck: I don't think golangci-lint needs to compile anything to run its checks, and probably doesn't share a lot of dependencies with our ship-it code. So I think we could split the stages like `Lint` and `Test and Build`, the advantage being that if the `Lint` stage fails travis won't even attempt to test and build.

tomtomyan: I think it would be even faster if we put everything (Lint, Test, Build) in 1 stage, in separate jobs. I checked our subscription and it looks like we are able to run 10 concurrent jobs. Let me know what you think.

crdueck: IMO I would stick to having the separate Lint and Test & Build stages. The Lint stage is the fastest stage, so there's quick PR feedback and since Travis runs the stages sequentially we can avoid testing/building code that doesn't pass lint.

A real downside is not seeing test failures if the lint stage fails, I think devs are/should be running their unit tests locally when they're changing code, but not everyone runs `golangci-lint` and with the exact same checks travis is running. So I'd rather have faster lint feedback even if it means sometime we don't let devs know about test failures as early as possible.

PR #133, opened 2019-12-03
Ship-it-api: provide the full tiller address

crdueck:


cmd/ship-it-api/main.go
```diff
@@ -10,12 +10,13 @@ import (
 
 	"ship-it/internal/api"
 	"ship-it/internal/api/config"
-	"ship-it/internal/api/integrations/helm"
+	helmstatus "ship-it/internal/api/integrations/helm"
```
crdueck: super awkward when we import both the real `helm` and our own `helm` pkgs...

PR #134, opened 2019-12-04
Add slack notifications

jrhouston:
VELO-1796


operator/notifications/notifier.go
```diff
@@ -0,0 +1,8 @@
+// Notifier
+
+package notifications
+
+// Notifier sends a notification
+type Notifier interface {
```
crdueck: Interfaces should be defined where they are used, rather than where they are implemented. 

https://github.com/Wattpad/highlander/wiki/Go-best-practices:--avoiding-%22Java-style%22-interfaces

operator/main.go
```diff
@@ -65,6 +69,9 @@ func main() {
 	flag.DurationVar(&gracePeriod, "grace-period", 10*time.Second, "The duration the operator will wait before checking a release's status after reconciling")
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
+	flag.StringVar(&slackChannel, "slack-channel", "us-east-1", "The channel to send Slack notifications to")
```
crdueck: I don't think `us-east-1` is a good default value here, probably some leftover copypaste

operator/controllers/helmrelease_controller_test.go
```diff
@@ -22,6 +22,13 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
+type fakeSlack struct{}
+
+func (m *fakeSlack) Send(msg string) error {
+	fmt.Println(msg)
```
crdueck: Printing the messages to stdout makes it difficult to programmatically inspect the sent messages. We could make `fakeSlack` easier to test by removing the IO side-effect and accumulating a `[]string` instead.

operator/notifications/slack/slack.go
```diff
@@ -0,0 +1,26 @@
+package slack
```
crdueck: We've named a bunch of ship-it pkgs to collide with the "official" pkgs. It's ended up causing awkward imports whenever we're forced to import both in the same file.

I'd suggest organizing this code like `notifications/slack.go` and then have 

```
package notifications

type Slack struct { ... }
```

Which results in an obviously meaningful imported type `notifications.Slack`, whereas I don't think `slack.Manager` is as meaningful.

operator/controllers/helmrelease_controller.go
```diff
@@ -294,6 +301,7 @@ func (r *HelmReleaseReconciler) rollback(ctx context.Context, rls *shipitv1beta1
 	}
 
 	r.Log.Info("rolling back HelmRelease", "release", releaseName)
+	r.notifier.Send(fmt.Sprintf("⚠️ `%s` is being rolled back.", releaseName))
```
crdueck: If `manager.Rollback` succeeds, but then the `Status().Update(...)` fails, we wouldn't notify slack. Same for all the other install/upgrade/delete etc. actions.

Not sure if it actually matters or if there's a proper solution that doesn't involve transactions 🙀, but we could avoid that scenario if slack notifications happened closer to the helm client calls: https://github.com/Wattpad/ship-it/blob/master/operator/controllers/release_manager.go#L16

operator/main.go
jrhouston: oops, yep

deploy/ship-it/templates/shipit.wattpad.com_helmreleases.yaml
```diff
@@ -432,7 +434,10 @@ spec:
               type: array
           type: object
       type: object
-  version: v1beta1
+  versions:
```
jrhouston: This got regenerated when I ran tests locally.

operator/controllers/helmrelease_controller.go
jrhouston: Yeah I thought about this but I think this is OK for now. If updating the status of a custom resource fails than that's likely to be a problem with kube itself and not something the developer is going to be able to action (other than to tell us). We'll probably want to spit out a log message at ERROR level and have an alert for it so velocity can investigate.

PR #135, opened 2019-12-05
Fix make kind-up errors

tomtomyan:


Makefile
```diff
@@ -45,7 +45,7 @@ docs: api/*.json docs/operator-release-states.png
 
 kind-up:
 	@echo Creating the $(KIND_CLUSTER_NAME) cluster...
-	kind create cluster --config hack/$(KIND_CLUSTER_NAME).yaml --name $(KIND_CLUSTER_NAME) >/dev/null 2>&1 || true
```
crdueck: We should keep the `|| true` so that this command is idempotent.

PR #138, opened 2019-12-09
Handle "no affected registered releases" edge case in ECR reconciler

crdueck:
If a new image is pushed but there are no registered releases that use
the image, the chart editor can exit early with a meaningful error
message. The error message can be logged so we can see what syncd is
doing even when there aren't many services registered.

---

I'm realizing there's a bit of ugliness in the way ship-it tracks the state of the registry chart right now. Most of the time the state of the registry chart release in the cluster is the source of truth, but this logic uses the state of the registry chart in github. 

Though those 2 states will eventually converge, it's possible that they differ at the moment ship-it observes them. I can't think of a concrete case where this actually causes problems, but I'm probably not being creative enough.

### DIFF
--- a/internal/syncd/integrations/ecr/chart_editor_test.go
+++ b/internal/syncd/integrations/ecr/chart_editor_test.go
@@ -2,6 +2,7 @@ package ecr
 
 import (
 	"context"
+	"encoding/base64"
 	"path"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


internal/syncd/integrations/ecr/chart_editor_test.go
```diff
@@ -151,9 +152,10 @@ spec:
 
 	// expect GetBlob per changed file
 	for i, entry := range modifiedTreeEntries {
+		blob64 := base64.StdEncoding.EncodeToString(blobContents[i])
```
crdueck: This test was kinda broken before, but I didn't notice >.<

internal/syncd/integrations/ecr/reconciler_test.go
```diff
@@ -107,3 +108,30 @@ func TestReconcilerSuccess(t *testing.T) {
...
+			Name:      "bar",
+		},
+	}
+
+	mockIndexer.On("Lookup", inputImage).Return(releaseNames, error(nil))
+
+	mockEditor.On("Edit", mock.Anything, releaseNames, inputImage).Return(errors.Wrap(errNoRegisteredReleasesAffected, "extra context"))
+
+	err := reconciler.Reconcile(context.Background(), inputImage)
+	assert.NoError(t, err)
```
akdh: The expectation is that this is an error correct?

crdueck: When the editor returns `errNoRegisteredReleasesAffected`, the reconciler should log the error but return `nil` because it successfully reconciled 0 releases. So the `NoError` assertion is correct here.

PR #142, opened 2019-12-10
Fix long service names wrap and break the grid

jrhouston:
Screen Shot 2019-12-10 at 3 41 03 PM


jrhouston:
screenie

Added a tooltip

PR #143, opened 2019-12-10
Fix helm status not displaying

jrhouston:
Screen Shot 2019-12-10 at 4 49 50 PM

### DIFF
--- a/web/src/components/ExpandedCard.js
+++ b/web/src/components/ExpandedCard.js
@@ -29,7 +29,7 @@ class ExpandedCard extends React.Component {
     podClick = () => {
         this.setState({ podsVisible: !this.state.podsVisible })
         axios.get(urljoin(this.props.API_ADDRESS, 'releases', this.props.data.name, 'resources')).then(response => {
-            this.setState({ resourceString: response.data.resources })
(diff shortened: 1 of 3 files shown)


web/src/components/ExpandedCard.js
```diff
@@ -29,7 +29,7 @@ class ExpandedCard extends React.Component {
     podClick = () => {
         this.setState({ podsVisible: !this.state.podsVisible })
         axios.get(urljoin(this.props.API_ADDRESS, 'releases', this.props.data.name, 'resources')).then(response => {
-            this.setState({ resourceString: response.data.resources })
+            this.setState({ resourceString: response.data.status })
```
crdueck: d'oh

web/src/components/GridCell.js
```diff
@@ -83,9 +84,9 @@ class SingleGridCell extends React.Component {
               </Typography>
               <StatusChip status={this.props.SingleGridCellData.status} />
               <div>
-                <IconButton>
+                <Icon>
```
tomtomyan: The alignment is awkward but we are probably going to change it later when we add in the time. LGTM.

